PR #5, opened 2014-08-22
 Media information and tags support.

ajaysusarla:
This patch adds support to retrieve media information for
audio/video/text streams.

Tags from each individual stream for Audio. Video and Text
 are parsed and stored in the appropriate structure 
(GstPlayerAudioStreamInfo, GstPlayerVideoStreamInfo, and
 GstPlayerTextStreamInfo respectively).

The primary structure is GstPlayerMediaInfo. An instance of
 this structure for each stream type is in GstPlayer.

Once the data is parsed from each of these streams, data
 is passed to the client via the signals 
(AUDIO_DATA_UPDATED, VIDEO_DATA_UPDATED,
 TEXT_DATA_UPDATED).


sdroege:
This is obsoleted by https://github.com/sdroege/gst-player/pull/9

PR #10, opened 2015-03-03
Android: cast pointers to gintptr

xclaesse:


sdroege:
Thanks, merged :smile:

PR #11, opened 2015-03-11
Add a 'missing-plugin' signal to player

jku:
Also use the signal to notify user in both gst-play and gtk-play.

For gtk-play this means a GtkInfoBar that says e.g. `Missing plugin 'H.264 decoder'.` For gst-play it's just a g_warning() with same content.

### DIFF
--- a/gtk/gtk-play.c
+++ b/gtk/gtk-play.c
@@ -50,6 +50,8 @@ typedef struct
   GtkWidget *prev_button, *next_button;
   GtkWidget *seekbar;
   GtkWidget *video_area;
+  GtkWidget *info_label;
+  GtkWidget *info_bar;
   GtkWidget *volume_button;
   gulong seekbar_value_changed_signal_id;
   gboolean playing;
@@ -120,6 +122,13 @@ play_pause_clicked_cb (GtkButton * button, GtkPlay * play)
 }
 
 static void
+clear_missing_plugins (GtkPlay * play)
+{
+  gtk_label_set_text (GTK_LABEL (play->info_label), "");
+  gtk_widget_hide (play->info_bar);
+}
+
+static void
 skip_prev_clicked_cb (GtkButton * button, GtkPlay * play)
 {
   GList *prev;
@@ -134,6 +143,7 @@ skip_prev_clicked_cb (GtkButton * button, GtkPlay * play)
 
   gtk_widget_set_sensitive (play->next_button, TRUE);
   gst_player_set_uri (play->player, prev->data);
+  clear_missing_plugins (play);
   gst_player_play (play->player);
   set_title (play, prev->data);
   gtk_widget_set_sensitive (play->prev_button, g_list_previous (prev) != NULL);
@@ -154,6 +164,7 @@ skip_next_clicked_cb (GtkButton * button, GtkPlay * play)
 
   gtk_widget_set_sensitive (play->prev_button, TRUE);
   gst_player_set_uri (play->player, next->data);
+  clear_missing_plugins (play);
   gst_player_play (play->player);
   set_title (play, next->data);
   gtk_widget_set_sensitive (play->next_button, g_list_next (next) != NULL);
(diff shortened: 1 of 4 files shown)


sdroege:
Thanks for the patch but I'm not sure how useful that is in general. Wouldn't an application prefer some API that directly lets you install the plugins?

jku:
Yeah, I should have talked about that... The gst API for this exists for two reasons : 
1. for concise error/problem reporting to the user mentioning what exactly is missing
2. for initiating installation of missing plugins

and I only covered part 1 here because I wasn't going to do the actual installation myself anyway... but if you prefer that, I could include "installer_detail" string in the signal, so applications could then use gst_install_plugins_async () with that...

sdroege:
IMHO ideally 1) should be solved by putting these information into the error or warning message that is posted after the missing-plugin messages, instead of providing a second API for a different kind of errors.

jku:
I'm afraid I don't understand waht you mean by "a second API for different kind of errors".  I'll try to explain my suggestion more clearly:

So the "installer_detail" I refer to would be from gst_missing_plugin_message_get_installer_detail() and is not really human readable (not for normal people anyway). So I think the best we can do for issue 1 is what I already provided. But the installer_detail might indeed be useful for applications that do want to start a plugin installation process to fix issue 2: I was proposing adding that string to the signal that is already in the patch.

sdroege:
For 1), what we have already should be enough. There should be an error message that tells exactly what is missing, and if not that should be fixed :smile: With a second API I meant the "error" signal as the primary error API, and your new "missing-plugin" as a secondary API.

For the installer stuff, I would prefer to abstract that a bit more so that people wouldn't have to use magic GStreamer API but that it would just do the right thing easily, and install plugins if needed and if allowed by the application. Without leaking the non-human readable details about things to the application.

jku:
Oh, now I finally got what you meant. Let me try the error signal -- it should indeed be usable for this.

jku:
How about now? It uses "error" signal to achieve the same UX result.

If you don't think this is worthwhile, that's fine too: but I'm happy to keep iterating if I can make this into something you'd like to take into upstream gst-player...

sdroege:
Thanks, I'll take a look later.

I definitely want to integrate something for this, but I'm also not completely sure yet about what API would be useful to have :)

lib/gst/player/gstplayer.c
```diff
@@ -656,6 +658,21 @@ error_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  GstPlayer *self = GST_PLAYER (user_data);
+
+  if (gst_is_missing_plugin_message (msg)) {
+    gchar *desc;
+
+    desc = gst_missing_plugin_message_get_description (msg);
+    emit_error (self, g_error_new (GST_PLAYER_ERROR,
+        GST_PLAYER_ERROR_MISSING_PLUGIN, "Missing plugin '%s'", desc));
+    g_free (desc);
+  }
```
sdroege: This is almost ok, but note that there can be multiple missing-plugin messages in theory. And only after the last one an error message is posted, or a warning message if only irrelevant plugins are missing and playback can continue.

I think you should accumulate all missing plugin messages and wait until the error or warning message (GST_CORE_ERROR_MISSING_PLUGIN or GST_STREAM_ERROR_CODEC_NOT_FOUND). And only if it's an error message, actually emit an error. As emitting an error will stop playback, which might not be needed.
In case of a warning message, it might make sense to add a warning signal that mirrors the error signal but has the big difference that it does not cause playback to stop.

gtk/gtk-play.c
```diff
@@ -316,6 +349,24 @@ eos_cb (GstPlayer * unused, GtkPlay * play)
...
 
+static void
+error_cb (GstPlayer * player, GError * err, GtkPlay * play)
+{
+  char *message;
+
+  if (g_error_matches (err, gst_player_error_quark (),
+      GST_PLAYER_ERROR_MISSING_PLUGIN)) {
+    // add message to end of any existing message: there may be
+    // multiple missing plugins.
```
sdroege: Use C-style comments please :) /* ... */

gtk/gtk-play.c
```diff
@@ -316,6 +349,24 @@ eos_cb (GstPlayer * unused, GtkPlay * play)
   }
 }
 
+static void
+error_cb (GstPlayer * player, GError * err, GtkPlay * play)
+{
+  char *message;
+
+  if (g_error_matches (err, gst_player_error_quark (),
+      GST_PLAYER_ERROR_MISSING_PLUGIN)) {
```
sdroege: I think it also makes sense to show other errors in the UI

sdroege:
See comments inside the commit, this is definitely going into the right direction now. Thanks for working on it :smile:

sdroege:
Are you planning to continue working on this?

jku:
Yes, I've just been distracted -- I think I can look at it later next week... (but feel free to just take what's there and modify if you want)

sdroege:
Thanks, it's not urgent :)

rossburton:
The commit that "closes" this didn't actually close it, but had a stack track in so closed all bugs 1 through 14.  Reopen please!

jku:
I notice I've promised things in this bug that I've not delivered on :/ 

I won't do anything about it in the next weeks. I'll try to fit it in the calendar later but won't mind if someone else grabs it.

sdroege:
https://bugzilla.gnome.org/show_bug.cgi?id=765934

PR #13, opened 2015-03-29
add media information and tag list support

codomania:
Signed-off-by: Brijesh Singh <brijesh.ksingh@gmail.com>

### DIFF
--- a/lib/gst/player/gstplayer-media-info-private.h
+++ b/lib/gst/player/gstplayer-media-info-private.h
@@ -76,11 +76,6 @@ struct _GstPlayerMediaInfo
 
   gchar *uri;
 
-  /* current streams */
-  GstPlayerStreamInfo  *current_audio;
-  GstPlayerStreamInfo  *current_video;
-  GstPlayerStreamInfo  *current_subtitle;
-
   /* all the streams */
   GList *stream_list;
 
--- a/lib/gst/player/gstplayer-media-info-private.h
+++ b/lib/gst/player/gstplayer-media-info-private.h
@@ -1,0 +1,89 @@
+/* GStreamer
+ * Copyright (C) 2015 Brijes Singh <brijesh.ksingh@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin St, Fifth Floor,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "gstplayer-media-info.h"
+
+struct _GstPlayerStreamInfo
+{
+  GObject parent;
+
+  GstCaps *caps;
+  gint stream_id;
+  GstTagList  *tags;
+
+  gpointer _gst_reserved[GST_PADDING];
+};
+
+struct _GstPlayerSubtitleInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  gchar *language;
+
+  gpointer _gst_reserved[GST_PADDING];
+};
+
+struct _GstPlayerAudioInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  guint channels;
+  guint sample_rate;
+  guint depth;
+
+  guint bitrate;
+  guint max_bitrate;
+
+  gchar *language;
+
+  gpointer _gst_reserved[GST_PADDING];
+};
+
+struct _GstPlayerVideoInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  guint width;
+  guint height;
+  guint framerate_num;
+  guint framerate_denom;
+
+  guint bitrate;
+  guint max_bitrate;
+
+  gpointer _gst_reserved[GST_PADDING];
+};
+
+struct _GstPlayerMediaInfo
+{
+  GObject parent;
+
+  gchar *uri;
+
+  /* current streams */
+  GstPlayerStreamInfo  *current_audio;
+  GstPlayerStreamInfo  *current_video;
+  GstPlayerStreamInfo  *current_subtitle;
+
+  /* all the streams */
+  GList *stream_list;
+
+  GstClockTime  duration;
+};
+
--- a/lib/gst/player/gstplayer-media-info.c
+++ b/lib/gst/player/gstplayer-media-info.c
@@ -513,7 +513,7 @@ G_DEFINE_TYPE (GstPlayerMediaInfo, gst_player_media_info, G_TYPE_OBJECT);
 static void
 gst_player_media_info_init (GstPlayerMediaInfo * info)
 {
-  /* nothing to do here */
+  info->duration = -1;
 }
 
 static void
@@ -613,5 +613,19 @@ gst_player_media_info_get_stream_list (const GstPlayerMediaInfo *info)
     list = g_list_append (list, l->data);
 
   return list;
+}
+
+/**
+ * gst_player_media_info_get_duration:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (GstClockTime) Duration of the media.
+ */
+GstClockTime
+gst_player_media_info_get_duration (const GstPlayerMediaInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), -1);
+
+  return info->duration;
 }
 
--- a/lib/gst/player/gstplayer-media-info.c
+++ b/lib/gst/player/gstplayer-media-info.c
@@ -1,0 +1,617 @@
+#include "gstplayer-media-info.h"
+#include "gstplayer-media-info-private.h"
+
+/* Per-stream information */
+G_DEFINE_TYPE (GstPlayerStreamInfo, gst_player_stream_info, G_TYPE_OBJECT);
+
+static void
+gst_player_stream_info_init (GstPlayerStreamInfo * sinfo)
+{
+  /* Nothing to do here */
+}
+
+static void
+gst_player_stream_info_finalize (GObject * object)
+{
+  GstPlayerStreamInfo *sinfo = GST_PLAYER_STREAM_INFO (object);
+
+  if (sinfo->caps) {
+    gst_caps_unref (sinfo->caps);
+    sinfo->caps = NULL;
+  }
+
+  if (sinfo->tags) {
+    gst_tag_list_unref (sinfo->tags);
+    sinfo->tags = NULL;
+  }
+}
+
+static void
+gst_player_stream_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_stream_info_finalize;
+}
+
+/**
+ * gst_player_stream_info_get_stream_id:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the stream ID of this stream.
+ */
+gint
+gst_player_stream_info_get_stream_id (const GstPlayerStreamInfo *info)
+{
+  return info->stream_id;
+}
+
+/**
+ * gst_player_stream_info_get_stream_type_nick:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: a human readable name for the stream type of the given @info (ex : "audio",
+ * "video",...).
+ */
+const gchar*
+gst_player_stream_info_get_stream_type_nick (const GstPlayerStreamInfo *info)
+{
+  if (GST_IS_PLAYER_VIDEO_INFO (info))
+    return "video";
+
+  if (GST_IS_PLAYER_AUDIO_INFO (info))
+    return "audio";
+
+  if (GST_IS_PLAYER_SUBTITLE_INFO (info))
+    return "subtitle";
+
+  return NULL;
+}
+
+/**
+ * gst_player_stream_info_get_tags:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the tags contained in this stream. Unref with #gst_tag_list_unref
+ */
+GstTagList*
+gst_player_stream_info_get_stream_tags (const GstPlayerStreamInfo *sinfo)
+{
+  GstTagList *tags = NULL;
+
+  if (sinfo->tags)
+    tags = gst_tag_list_ref (sinfo->tags);
+
+  return tags;
+}
+
+/**
+ * gst_player_stream_info_get_caps:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the #GstCaps of the stream. Unref with #gst_caps_unref after usage.
+ */
+GstCaps*
+gst_player_stream_info_get_stream_caps (const GstPlayerStreamInfo *sinfo)
+{
+  GstCaps *caps = NULL;
+
+  if (sinfo->caps)
+    caps = gst_caps_ref (sinfo->caps);
+
+  return caps;
+}
+
+/**
+ * gst_player_video_info_get_streams:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (transfer full) (element-type GstPlayerVideoInfo): A #GList of
+ * matching #GstPlayerVideoInfo. The caller should free it with
+ * gst_player_stream_info_list_free().
+ */
+GList*
+gst_player_media_info_get_video_streams (const GstPlayerMediaInfo *info)
+{
+  GList *result = NULL, *p;
+
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  for (p = info->stream_list; p != NULL; p = p->next) {
+    GstPlayerStreamInfo  *sinfo = (GstPlayerStreamInfo*) p->data;
+
+    if (GST_IS_PLAYER_VIDEO_INFO (sinfo)) {
+      GstPlayerVideoInfo  *vinfo = (GstPlayerVideoInfo*) sinfo;
+
+      result = g_list_append (result, vinfo);
+    }
+  }
+
+  return result;
+}
+
+/**
+ * gst_player_subtitle_info_get_streams:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (transfer full) (element-type GstPlayerSubtitleInfo): A #GList of
+ * matching #GstPlayerSubtitleInfo. The caller should free it with
+ * gst_player_stream_info_list_free().
+ */
+GList*
+gst_player_media_info_get_subtitle_streams (const GstPlayerMediaInfo *info)
+{
+  GList *result = NULL, *p;
+
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  for (p = info->stream_list; p != NULL; p = p->next) {
+    GstPlayerStreamInfo  *sinfo = (GstPlayerStreamInfo*) p->data;
+
+    if (GST_IS_PLAYER_SUBTITLE_INFO (sinfo)) {
+      GstPlayerSubtitleInfo  *tinfo = (GstPlayerSubtitleInfo*) sinfo;
+
+      result = g_list_append (result, tinfo);
+    }
+  }
+
+  return result;
+}
+
+/**
+ * gst_player_audio_info_get_streams:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (transfer full) (element-type GstPlayerAudioInfo): A #GList of
+ * matching #GstPlayerAudioInfo. The caller should free it with
+ * gst_player_stream_info_list_free().
+ */
+GList*
+gst_player_media_info_get_audio_streams (const GstPlayerMediaInfo *info)
+{
+  GList *result = NULL, *p;
+
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  for (p = info->stream_list; p != NULL; p = p->next) {
+    GstPlayerStreamInfo  *sinfo = (GstPlayerStreamInfo*) p->data;
+
+    if (GST_IS_PLAYER_AUDIO_INFO (sinfo)) {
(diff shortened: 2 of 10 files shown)


codomania:
Please review the patch. The media info support has been validated using the below command
./gst-play/gst-play --interactive http://trailers.divx.com/divx_prod/divx_plus_hd_showcase/Sintel_DivXPlus_6500kbps.mkv

Press 'i' to query the media information during the playback.

codomania:
I have add track selection feature in gtk-play. 

image

codomania:
the above commit adds track selection support in gtk-play ...

image

lib/gst/player/gstplayer-media-info-private.h
```diff
@@ -0,0 +1,84 @@
...
+
+  gpointer _gst_reserved[GST_PADDING];
+};
+
+struct _GstPlayerVideoInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  guint width;
+  guint height;
```
sdroege: Will this contain the display width/height, or the pixels? In case of the latter you also should include the pixel-aspect-ratio here

lib/gst/player/gstplayer-media-info-private.h
```diff
@@ -0,0 +1,84 @@
...
+
+struct _GstPlayerStreamInfo
+{
+  GObject parent;
+
+  GstCaps *caps;
+  gint stream_id;
+  GstTagList  *tags;
+
+  gpointer _gst_reserved[GST_PADDING];
```
sdroege: Padding is only really required for public API, which this is not

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+/**
+ * gst_player_stream_info_get_stream_id:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the stream ID of this stream.
+ */
+gint
+gst_player_stream_info_get_stream_id (const GstPlayerStreamInfo *info)
+{
+  return info->stream_id;
```
sdroege: It might be better to call this "stream index", and for stream-id you could use the string-typed stream-id from the -stream-start event of the relevant stream. See gst_event_parse_stream_start()

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+    return "subtitle";
+
+  return NULL;
+}
+
+/**
+ * gst_player_stream_info_get_tags:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the tags contained in this stream. Unref with #gst_tag_list_unref
```
sdroege: (transfer full) annotation needed here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+    tags = gst_tag_list_ref (sinfo->tags);
+
+  return tags;
+}
+
+/**
+ * gst_player_stream_info_get_caps:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the #GstCaps of the stream. Unref with #gst_caps_unref after usage.
```
sdroege: (transfer full) annotation needed here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  GstCaps *caps = NULL;
+
+  if (sinfo->caps)
+    caps = gst_caps_ref (sinfo->caps);
+
+  return caps;
+}
+
+/**
+ * gst_player_video_info_get_streams:
```
sdroege: gst_player_get_video_streams() seems like a better name for this (same for audio/subs)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  for (p = info->stream_list; p != NULL; p = p->next) {
+    GstPlayerStreamInfo  *sinfo = (GstPlayerStreamInfo*) p->data;
+
+    if (GST_IS_PLAYER_VIDEO_INFO (sinfo)) {
+      GstPlayerVideoInfo  *vinfo = (GstPlayerVideoInfo*) sinfo;
+
+      result = g_list_append (result, vinfo);
```
sdroege: As you say this is transfer-full (which is the right thing to do), you need to store a new reference in the returned list.

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  return result;
+}
+
+/**
+ * gst_player_audio_info_get_language:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the language of the stream, or NULL if unknown.
+ */
+const gchar*
```
sdroege: Not const and needs to be (transfer full)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+/**
+ * gst_player_media_stream_info_list_free:
+ * @infos: a #GList of #GstPlayerStreamInfo
+ *
+ * frees the #GList.
+ */
+void
+gst_player_media_info_stream_info_list_free (GList *infos)
+{
+  g_list_free (infos);
```
sdroege: Needs to be g_list_free_full() with g_object_unref() as free function

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+}
+
+/**
+ * gst_player_media_stream_info_list_free:
+ * @infos: a #GList of #GstPlayerStreamInfo
+ *
+ * frees the #GList.
+ */
+void
+gst_player_media_info_stream_info_list_free (GList *infos)
```
sdroege: gst_player_stream_info_list_free()

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+}
+
+/**
+ * gst_player_video_info_get_framerate_num:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the framerate numerator of video in #GstPlayerVideoInfo.
+ */
+guint
+gst_player_video_info_get_framerate_num (const GstPlayerVideoInfo *info)
```
sdroege: void gst_player_video_info_get_framerate (const GstPlayerVideoInfo *info, guint *fps_n, guint *fps_d) would be better, and then making these out-parameters

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  /* Nothing to do here */
+}
+
+static void
+gst_player_audio_info_finalize (GObject * object)
+{
+  GstPlayerAudioInfo *ainfo = GST_PLAYER_AUDIO_INFO (object);
+
+  if (ainfo->language) {
+    g_free (ainfo->language);
```
sdroege: g_free() is NULL-safe

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  /* Nothing to do here */
+}
+
+static void
+gst_player_subtitle_info_finalize (GObject * object)
+{
+  GstPlayerSubtitleInfo *tinfo = GST_PLAYER_SUBTITLE_INFO (object);
+
+  if (tinfo->language) {
+    g_free (tinfo->language);
```
sdroege: g_free() is NULL-safe

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  klass->finalize = gst_player_subtitle_info_finalize;
+}
+
+/**
+ * gst_player_subtitle_info_get_language:
+ * @info: a #GstPlayerSubtitleInfo
+ *
+ * Returns: the language of the stream, or NULL if unknown.
+ */
+const gchar*
```
sdroege: Not const and should be (transfer full)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+ * Returns: the audio maximum bitrate in #GstPlayerAudioInfo.
+ */
+guint
+gst_player_audio_info_get_max_bitrate (const GstPlayerAudioInfo* info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_AUDIO_INFO (info), -1);
+
+  return info->max_bitrate;
+}
+
```
sdroege: There seems to be no getter for the audio language

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+    info->uri = NULL;
+  }
+
+  if (info->stream_list) {
+    GList *list;
+
+    for (list = info->stream_list; list != NULL; list = list->next)
+      g_object_unref (list->data);
+
+    g_list_free (info->stream_list);
```
sdroege: You can just use g_list_free_full() here instead of iterating over it yourself

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  info->duration = -1;
+}
+
+static void
+gst_player_media_info_finalize (GObject * object)
+{
+  GstPlayerMediaInfo  *info = GST_PLAYER_MEDIA_INFO (object);
+
+  if (info->uri) {
+    g_free (info->uri);
```
sdroege: g_free() is NULL-safe

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+  return NULL;
+}
+
+/**
+ * gst_player_media_info_get_uri:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: the URI associated with #GstPlayerMediaInfo.
+ */
+const gchar *
```
sdroege: Should not be const, return a copy of the string and (transfer full)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,562 @@
...
+ */
+GList *
+gst_player_media_info_get_stream_list (const GstPlayerMediaInfo *info)
+{
+  GList *list = NULL, *l;;
+
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  for (l = info->stream_list; l != NULL; l = l->next)
+    list = g_list_append (list, l->data);
```
sdroege: Should put a new reference in the list

lib/gst/player/gstplayer-media-info.h
```diff
@@ -0,0 +1,119 @@
...
+#define GST_IS_PLAYER_MEDIA_INFO(obj) \
+  (G_TYPE_CHECK_INSTANCE_TYPE((obj),GST_TYPE_PLAYER_MEDIA_INFO))
+#define GST_IS_PLAYER_MEDIA_INFO_CLASS(klass) \
+  (G_TYPE_CHECK_CLASS_TYPE((klass),GST_TYPE_PLAYER_MEDIA_INFO))
+
+typedef struct _GstPlayerMediaInfo GstPlayerMediaInfo;
+typedef GObjectClass GstPlayerMediaInfoClass;
+GType gst_player_media_info_get_type (void);
+
+GstPlayerMediaInfo* gst_player_media_info_new (const gchar *uri);
```
sdroege: The _new() function should be private, only GstPlayer is ever going to call it and not the application

lib/gst/player/gstplayer.h
```diff
@@ -109,6 +110,28 @@ void         gst_player_set_window_handle             (GstPlayer    * player,
...
+gboolean     gst_player_video_track_enable            (GstPlayer *player);
+
+gboolean     gst_player_subtitle_track_enable         (GstPlayer *player);
+
+GstPlayerAudioInfo * gst_player_get_audio_track       (GstPlayer *player);
+
+GstPlayerVideoInfo * gst_player_get_video_track       (GstPlayer *player);
+
+GstPlayerSubtitleInfo * gst_player_get_subtitle_track (GstPlayer *player);
+
```
sdroege: A gst_player_get_media_info() seems useful here too :)

lib/gst/player/player-media-info.h
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin St, Fifth Floor,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#ifndef __PLAYER_MEDIA_INFO_H__
+#define __PLAYER_MEDIA_INFO_H__
+
+#include <gst/player/gstplayer-media-info.h>
```
sdroege: This header should go away, and the #include above should go into player.h

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1149,594 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+{
+  gint current;
+  GstTagList  *tags;
+  GstPlayer *self = GST_PLAYER (user_data);
+
+  if (!self->priv->media_info)
+    return;
+
+  g_object_get (G_OBJECT (self->priv->playbin), "current-text",
+                &current, NULL);
```
sdroege: You don't seem to do anything with "current" here

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1149,594 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    return;
+
+  g_object_get (G_OBJECT (self->priv->playbin), "n-text", &n_text,
+                  "current-text", &current, NULL);
+
+  GST_DEBUG_OBJECT (self, "n-text: %d, current: %d", n_text, current);
+
+  for (i = 0; i < n_text; i++) {
+    g_signal_emit_by_name (self->priv->playbin, "get-text-tags", i, &tags);
+    update_stream_info (self, i, GST_TYPE_PLAYER_SUBTITLE_INFO, tags);
```
sdroege: Note that also the *number* of streams can change here

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1149,594 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    /* create a new stream object */
+    stream_info = stream_info_new (self, type, stream_id);
+
+    if (stream_info == NULL) {
+      return NULL;
+    }
+  }
+
+  result = gst_tag_list_merge (stream_info->tags, tags,
+                               GST_TAG_MERGE_REPLACE);
```
sdroege: I think the taglist needs to be completely replaced. What was stream X before, might now be a completely different stream and have nothing to do at all anymore with the previous one

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1149,594 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  if (type == GST_TYPE_PLAYER_SUBTITLE_INFO)
+    g_signal_emit_by_name (G_OBJECT (self->priv->playbin),
+                          "get-text-pad", stream_id, &pad);
+
+  if (pad)
+    caps = gst_pad_get_current_caps (pad);
+
+  if (caps) {
+    outcaps = gst_caps_copy (caps);
+    gst_caps_unref (caps);
```
sdroege: No need to copy, just return those caps instead of unreffing them

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1149,594 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+  return ret;
+}
+
+static GstPlayerStreamInfo *
+gst_player_get_current_stream (GstPlayer *self, GType type)
+{
+  guint flag;
+  gint stream_id;
+  gchar *prop = NULL;
```
sdroege: Make this const, and don't use g_strdup() below

lib/gst/player/gstplayer.c
```diff
@@ -253,6 +270,9 @@ gst_player_finalize (GObject * object)
 
   GST_TRACE_OBJECT (self, "Finalizing");
 
+  if (self->priv->media_info)
+    g_object_unref (self->priv->media_info);
```
sdroege: The media_info should also always be destroyed when a the player is stopped / or a new URI is selected

sdroege:
Sorry for the late review. This looks generally very good already :) Just some minor comments above

One major thing should be changed though, which is not in the comments above. IMHO the GstPlayerMediaInfo (and thus all things contained inside it) should be immutable. That is, after you created and filled it you will never change it again. Instead whenever something changes, you create a new GstPlayerMediaInfo (and everything contained inside it).

That way an application can be sure that whenever it has a GstPlayerMediaInfo (or something contained inside it), that it will never change under its fingers unexpectedly. But that it might get a new signal with a new info at some point.

(Then you also can return the strings/taglists/etc directly from the stream infos instead of creating copies or new references)

sdroege:
Please let me know if you have any questions, and also if you don't have the time to continue working on this :)

codomania:
Thanks for review comment. I will look at each of the comments and address those. 

I like your suggestion about making the mediainfo immutable but just wanted to get your thought a bit more before I go ahead and implement it and test.

To my knowledge the this is how playbin works and I am trying to visualize when and where we should make a new mediainfo object and emit signal. This is how the current implementation looks like:

1) create new mediainfo object when uri is set.

2) whenever a new stream is detected then video/audio/text-changed callback is called. In this callback we  create a new streaminfo object (video or audio or subtitle) and add this into media info stream list.

3) whenever a tag is changed then video/audio/text-tag-changed callback is executed and in this call we try to find the streaminfo object corresponding to stream id and update tags, caps and other fields of that particular stream object. This when we emit "media-info-updated" signal to user app. From this point onwards userapp uses the mediainfo object to get all those media information.

You are right is background if audio/video/text-tag-changed or audio/video/text-changed callback is called by playbin then we add/update the stream information. 

If we like to make the mediainfo immutable then where do you think we should handle it ? 
One simple approach is create a new media-info-object and copy the current media-info object content into newly created object and return this newly created mediainfo object from "media-info-updated" signal. 

BTW, regarding your question on time, I would like to finish this and get it merged and will  try to keep working in my after work hrs.
In my branch, I have full code working just fine. Not sure if you have seen but I have made all those gtk changes so that you can select different tracks and also enable/disable track. Otherwise than new suburi I have pretty much everything implemented.

codomania:
In last line, I was meaning to say "Other than new suburi I have pretty much everything implemented .."

lib/gst/player/gstplayer-media-info-private.h
codomania: Yes. Will add pixel-aspect-ratio.

lib/gst/player/gstplayer-media-info-private.h
codomania: okay thanks

lib/gst/player/gstplayer-media-info.c
codomania: okay

lib/gst/player/gstplayer-media-info.c
codomania: okay

lib/gst/player/gstplayer-media-info.c
codomania: okay

lib/gst/player/gstplayer-media-info.c
codomania: okay

lib/gst/player/gstplayer-media-info.c
codomania: okay

lib/gst/player/gstplayer-media-info.c
codomania: okay

sdroege:
Yeah, I saw your GTK changes. Looks great :) Thanks for working on that.

For the audio/video/text-changed signals, they are not just when a new stream is detected. It might also be a stream removed. Basically you have to recreate all streams of that type whenever that signal is emitted.


Personally I would just create a completely new GstPlayerMediaInfo + all the stream infos from scratch whenever any of those signals is emitted (and destroy it when URI changes or player is stopped).

And whenever the *-tags-changed signal is emitted, I would create a new GstPlayerMediaInfo object that new references (not copies) to all the streams... and create a copy of the stream that has changed tags. And then update the tags only in the copy (and completely replace the taglist instead of merging with REPLACE. playbin is aggregating tags for us already).

Also what seems to be missing is that we need to install notify::caps signal handlers on all audio, video and text pads. And whenever the caps change, we would do the same as for the *-tags-changed signals.

Do you agree?

codomania:
Sounds good. Let me give a first pass on implementing it and send you for review comment. Since we will sending new GstPlayerMediaInfo object on our media-updated signal it sounds like I need to also do book keeping (may be GList) of the all the object send from our media-updated signal and all these need to be unref'ed.

sdroege:
What kind of book keeping? I don't think that is required. Also always store the very last media info inside GstPlayer (and add a getter for that) :)

codomania:
Here is what I am thinking:

- Add three list variables in GstPlayer->priv
 {
     ......
     GList *audioinfo_list, *videoinfo_list, *subinfo_list;
    ....
 }

* on *-changed signal, destroy the list and create a new stream objects for all the known stream_indexes and add them in the GList.
* on *-tags-changed or notify:caps, create a new GstPlayerMediaInfo object and add (by referencing) all the stream object found in GLists (audio, video and sub) into GstPlayerMediaInfo->stream_list. While adding the object if its current stream then instead of reference, create a new stream object and copy the contents, update the tag, update the caps and then add this new object in stream_list. Finally emit the "media-updated" signal to user.

As you can see user app will be getting multiple media-update signal and each signal will have a new GstPlayerMediaInfo object hence user app will be responsible to unref all these object else will leak the memory. And to avoid that I was hoping that creating a GList to store the GstPlayerMediaInfo objects created during the *-tags-changed and then we simply unref it in uri change or player stopped.

sdroege:
I would just store the very latest GstPlayerMediaInfo in GstPlayer. And then whenever something changes, you either create a completely new one or you create a shallow copy and then replace the relevant stream. No need to keep track of things separately.

The app is not responsible for unreffing it, instead you would keep the GstPlayerMediaInfo inside the GstPlayer and unref it when replacing it with a new one or a copy.


In the beginning it might also make sense to delay emission of the media-info-updated signal until we reached the PAUSED state (see the code for the initial time we go to PAUSED that is already in the state_changed_cb). That way we don't flood the application with intermediate, partial information.

sdroege:
New version in https://github.com/sdroege/gst-player/pull/14

PR #14, opened 2015-04-18
mediainfo-v1

codomania:
I have incorporated your review feedback in this commit. GstPlayerMediaInfo should be immutable now. In addition I check if pipeline prerolled state, if its prerolled then create a new GstPlayerMediaInfo object  (contain copy of all available streams) and emit this newly created GstPlayerMediaInfo object.

### DIFF
--- a/lib/gst/player/gstplayer-media-info-private.h
+++ b/lib/gst/player/gstplayer-media-info-private.h
@@ -1,0 +1,81 @@
+/* GStreamer
+ * Copyright (C) 2015 Brijesh Singh <brijesh.ksingh@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ * You should have received a copy of the GNU Library General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin St, Fifth Floor,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "gstplayer-media-info.h"
+
+struct _GstPlayerStreamInfo
+{
+  GObject parent;
+
+  GstCaps *caps;
+  gint stream_index;
+  GstTagList  *tags;
+};
+
+struct _GstPlayerSubtitleInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  gchar *language;
+};
+
+struct _GstPlayerAudioInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  guint channels;
+  guint sample_rate;
+  guint depth;
+
+  guint bitrate;
+  guint max_bitrate;
+
+  gchar *language;
+};
+
+struct _GstPlayerVideoInfo
+{
+  GstPlayerStreamInfo  parent;
+
+  guint width;
+  guint height;
+  guint framerate_num;
+  guint framerate_denom;
+  guint par_num;
+  guint par_denom;
+
+  guint bitrate;
+  guint max_bitrate;
+};
+
+struct _GstPlayerMediaInfo
+{
+  GObject parent;
+
+  gchar *uri;
+  GList *stream_list;
+
+  GstClockTime  duration;
+};
+
+GstPlayerStreamInfo*    gst_player_stream_info_new (GstPlayerStreamInfo *ref);
+GstPlayerVideoInfo*     gst_player_video_info_new (GstPlayerVideoInfo *ref);
+GstPlayerAudioInfo*     gst_player_audio_info_new (GstPlayerAudioInfo *ref);
+GstPlayerSubtitleInfo*  gst_player_subtitle_info_new (GstPlayerSubtitleInfo *ref);
+
--- a/lib/gst/player/gstplayer-media-info.c
+++ b/lib/gst/player/gstplayer-media-info.c
@@ -1,0 +1,603 @@
+#include "gstplayer-media-info.h"
+#include "gstplayer-media-info-private.h"
+
+/* Per-stream information */
+G_DEFINE_TYPE (GstPlayerStreamInfo, gst_player_stream_info, G_TYPE_OBJECT);
+
+static void
+gst_player_stream_info_init (GstPlayerStreamInfo * sinfo)
+{
+  sinfo->caps = NULL;
+  sinfo->tags = NULL;
+  sinfo->stream_index = -1;
+}
+
+static void
+gst_player_stream_info_finalize (GObject * object)
+{
+  GstPlayerStreamInfo *sinfo = GST_PLAYER_STREAM_INFO (object);
+
+  if (sinfo->caps) {
+    gst_caps_unref (sinfo->caps);
+    sinfo->caps = NULL;
+  }
+
+  if (sinfo->tags) {
+    gst_tag_list_unref (sinfo->tags);
+    sinfo->tags = NULL;
+  }
+
+  G_OBJECT_CLASS (gst_player_stream_info_parent_class)->finalize (object);
+}
+
+static void
+gst_player_stream_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_stream_info_finalize;
+}
+
+GstPlayerStreamInfo*
+gst_player_stream_info_new (GstPlayerStreamInfo *ref)
+{
+  GstPlayerStreamInfo *ret = NULL;
+
+  if (GST_IS_PLAYER_VIDEO_INFO (ref))
+    ret = (GstPlayerStreamInfo*) gst_player_video_info_new
+                                    ((GstPlayerVideoInfo*)ref);
+  if (GST_IS_PLAYER_AUDIO_INFO (ref))
+    ret = (GstPlayerStreamInfo*) gst_player_audio_info_new
+                                    ((GstPlayerAudioInfo*)ref);
+  if (GST_IS_PLAYER_SUBTITLE_INFO (ref))
+    ret = (GstPlayerStreamInfo*) gst_player_subtitle_info_new
+                                    ((GstPlayerSubtitleInfo*)ref);
+
+  ret->stream_index = gst_player_stream_info_get_stream_index (ref);
+  ret->tags = gst_player_stream_info_get_stream_tags (ref);
+  ret->caps = gst_player_stream_info_get_stream_caps (ref);
+
+  return ret;
+}
+
+/**
+ * gst_player_stream_info_get_stream_index:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the stream index of this stream.
+ */
+gint
+gst_player_stream_info_get_stream_index (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  return info->stream_index;
+}
+
+/**
+ * gst_player_stream_info_get_stream_type_nick:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: a human readable name for the stream type of the given @info (ex : "audio",
+ * "video",...).
+ */
+const gchar*
+gst_player_stream_info_get_stream_type_nick (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  if (GST_IS_PLAYER_VIDEO_INFO (info))
+    return "video";
+
+  if (GST_IS_PLAYER_AUDIO_INFO (info))
+    return "audio";
+
+  if (GST_IS_PLAYER_SUBTITLE_INFO (info))
+    return "subtitle";
+
+  return NULL;
+}
+
+/**
+ * gst_player_stream_info_get_tags:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: (transfer full): the tags contained in this stream.
+ */
+GstTagList*
+gst_player_stream_info_get_stream_tags (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  return info->tags;
+}
+
+/**
+ * gst_player_stream_info_get_caps:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: (transfer full): the #GstCaps of the stream.
+ */
+GstCaps*
+gst_player_stream_info_get_stream_caps (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  return info->caps;
+}
+
+/* Video information */
+G_DEFINE_TYPE (GstPlayerVideoInfo, gst_player_video_info,
+    GST_TYPE_PLAYER_STREAM_INFO);
+
+static void
+gst_player_video_info_init (GstPlayerVideoInfo * info)
+{
+  /* Nothing to do here */
+}
+
+static void
+gst_player_video_info_finalize (GObject * gobject)
+{
+  G_OBJECT_CLASS (gst_player_video_info_parent_class)->finalize (gobject);
+}
+
+static void
+gst_player_video_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_video_info_finalize;
+}
+
+GstPlayerVideoInfo*
+gst_player_video_info_new (GstPlayerVideoInfo *ref)
+{
+  GstPlayerVideoInfo *ret = NULL;
+
+  ret = g_object_new (GST_TYPE_PLAYER_VIDEO_INFO, NULL);
+
+  if (ref) {
+    ret->width = ref->width;
+    ret->height  = ref->height;
+    ret->framerate_num = ref->framerate_num;
+    ret->framerate_denom = ref->framerate_denom;
+    ret->par_num = ref->par_num;
+    ret->par_denom = ref->par_denom;
+    ret->bitrate = ref->bitrate;
+    ret->max_bitrate = ref->max_bitrate;
+  }
+
+  return ret;
+}
+
+/**
+ * gst_player_video_info_get_width:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the width of video in #GstPlayerVideoInfo.
+ */
+guint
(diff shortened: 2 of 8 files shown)


lib/gst/player/gstplayer-media-info-private.h
```diff
@@ -0,0 +1,81 @@
...
+  gchar *uri;
+  GList *stream_list;
+
+  GstClockTime  duration;
+};
+
+GstPlayerStreamInfo*    gst_player_stream_info_new (GstPlayerStreamInfo *ref);
+GstPlayerVideoInfo*     gst_player_video_info_new (GstPlayerVideoInfo *ref);
+GstPlayerAudioInfo*     gst_player_audio_info_new (GstPlayerAudioInfo *ref);
+GstPlayerSubtitleInfo*  gst_player_subtitle_info_new (GstPlayerSubtitleInfo *ref);
```
sdroege: Just for completeness, add a G_GNUC_INTERNAL here for all of them

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+    return "subtitle";
+
+  return NULL;
+}
+
+/**
+ * gst_player_stream_info_get_tags:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: (transfer full): the tags contained in this stream.
```
sdroege: This is actually (transfer none)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  return info->tags;
+}
+
+/**
+ * gst_player_stream_info_get_caps:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: (transfer full): the #GstCaps of the stream.
```
sdroege: (transfer none)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+static void
+gst_player_video_info_init (GstPlayerVideoInfo * info)
+{
+  /* Nothing to do here */
+}
+
+static void
+gst_player_video_info_finalize (GObject * gobject)
+{
+  G_OBJECT_CLASS (gst_player_video_info_parent_class)->finalize (gobject);
```
sdroege: Don't keep empty finalize implementations around

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+}
+
+static void
+gst_player_video_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_video_info_finalize;
+}
+
+GstPlayerVideoInfo*
+gst_player_video_info_new (GstPlayerVideoInfo *ref)
```
sdroege: I think the _new() methods should be split into a _new(void) and _copy(ref)

sdroege: Or maybe a new() that has all the fields the object contains

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+/**
+ * gst_player_video_info_get_framerate_num:
+ * @info: a #GstPlayerVideoInfo
+ *
+ */
+void
+gst_player_video_info_get_framerate (const GstPlayerVideoInfo *info,
+  guint *fps_n, guint *fps_d)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_VIDEO_INFO (info), NULL);
```
sdroege: g_return_if_fail(), you have void as return type here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+/**
+ * gst_player_video_info_get_pixel_aspect_ratio:
+ * @info: a #GstPlayerVideoInfo
+ *
+ */
+void
+gst_player_video_info_get_pixel_aspect_ratio (const GstPlayerVideoInfo *info,
+  guint *par_d, guint *par_n)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_VIDEO_INFO (info), NULL);
```
sdroege: g_return_if_fail(), you have void as return type here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+}
+
+/* Audio information */
+G_DEFINE_TYPE (GstPlayerAudioInfo, gst_player_audio_info,
+    GST_TYPE_PLAYER_STREAM_INFO);
+
+static void
+gst_player_audio_info_init (GstPlayerAudioInfo * info)
+{
+  info->language = NULL;
```
sdroege: It's automatically initialized to NULL, but maybe things like bitrate and samplerate and channels should be initialized to -1

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+  return ret;
+}
+
+/**
+ * gst_player_audio_info_get_language:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the language of the stream, or NULL if unknown.
+ */
+gchar*
```
sdroege: const

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+  info->language = NULL;
+}
+
+static void
+gst_player_subtitle_info_finalize (GObject * object)
+{
+  GstPlayerSubtitleInfo *info = GST_PLAYER_SUBTITLE_INFO (object);
+
+  g_free (info->language);
+
```
sdroege: info->language = NULL;

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+GstPlayerSubtitleInfo*
+gst_player_subtitle_info_new (GstPlayerSubtitleInfo *ref)
+{
+  GstPlayerSubtitleInfo *ret = NULL;
+
+  ret = g_object_new (GST_TYPE_PLAYER_SUBTITLE_INFO, NULL);
+
+  if (ref)
+    ret->language = g_strdup (ref->language);
```
sdroege: I think g_strdup(NULL) would crash, not 100% sure

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+  return ret;
+}
+
+/**
+ * gst_player_subtitle_get_language:
+ * @info: a #GstPlayerSubtitleInfo
+ *
+ * Returns: the language of the stream, or NULL if unknown.
+ */
+gchar*
```
sdroege: const

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+static void
+gst_player_media_info_finalize (GObject * object)
+{
+  GstPlayerMediaInfo  *info = GST_PLAYER_MEDIA_INFO (object);
+
+  g_free (info->uri);
+  info->uri = NULL;
+
+  if (info->stream_list) {
+    g_list_free (info->stream_list);
```
sdroege: g_list_free_full(info->stream_list, (GFunc) g_object_unref, NULL)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+  oclass->finalize = gst_player_media_info_finalize;
+}
+
+/**
+ * gst_player_media_info_get_uri:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: the URI associated with #GstPlayerMediaInfo.
+ */
+gchar *
```
sdroege: const

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  return info->uri;
+}
+
+/**
+ * gst_player_media_info_get_stream_list:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (transfer full) (element-type GstPlayerStreamInfo): A #GList of
```
sdroege: As we now made the GstPlayerMediaInfo immutable, we can make this (transfer none) and just directly return info->stream_list instead of copying it

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+    list = g_list_append (list, g_object_ref(l->data));
+
+  return list;
+}
+
+/**
+ * gst_player_get_video_streams:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (transfer full) (element-type GstPlayerVideoInfo): A #GList of
```
sdroege: Same here and for the others

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,603 @@
...
+}
+
+/* Global media information */
+G_DEFINE_TYPE (GstPlayerMediaInfo, gst_player_media_info, G_TYPE_OBJECT);
+
+static void
+gst_player_media_info_init (GstPlayerMediaInfo * info)
+{
+  info->uri = NULL;
+  info->stream_list = NULL;
```
sdroege: NULL not necessary here, but the duration should probably be initialized to -1. (Same for the audio/video integer fields probably, and par_n=par_d=1)

lib/gst/player/gstplayer.c
```diff
@@ -85,9 +87,17 @@ enum
   SIGNAL_END_OF_STREAM,
   SIGNAL_ERROR,
   SIGNAL_VIDEO_DIMENSIONS_CHANGED,
+  SIGNAL_MEDIA_INFO_UPDATED,
   SIGNAL_LAST
 };
 
+enum {
+  GST_PLAY_FLAG_VIDEO = (1UL << 0),
```
sdroege: I think these are just integers, so using 1 instead of 1UL seems more correct

lib/gst/player/gstplayer.c
```diff
@@ -111,6 +121,13 @@ struct _GstPlayerPrivate
 
   GstPlayerState app_state;
   gint buffering;
+  GList *audioinfo_list;
+  GList *videoinfo_list;
+  GList *subtitleinfo_list;
+
+  gboolean is_prerolled;
+  GstPlayerMediaInfo  *media_info;
+  GstPlayerMediaInfo  *prev_media_info;
```
sdroege: Only keep around the very latest one

lib/gst/player/gstplayer.c
```diff
@@ -111,6 +121,13 @@ struct _GstPlayerPrivate
 
   GstPlayerState app_state;
   gint buffering;
+  GList *audioinfo_list;
+  GList *videoinfo_list;
+  GList *subtitleinfo_list;
```
sdroege: Why? You should have these already in the media_info below

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  g_object_set (self->priv->playbin, "flags", &flags, NULL);
+}
+
+static gboolean
+is_track_enabled (GstPlayer *self, gint pos)
+{
+  gint flags;
+
+  g_object_get (self->priv->playbin, "flags", &flags, NULL);
+  if (flags & pos)
```
sdroege: I think this gives a compiler warning about missing parenthesis. Make this if ((flags & pos))

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+static void
+free_media_info_updated_signal_data (MediaInfoUpdatedSignalData *data)
+{
+  g_slice_free (MediaInfoUpdatedSignalData, data);
+}
+
+static void
+emit_media_updated_signal (GstPlayer *self, GstPlayerMediaInfo *info,
+  guint signal)
```
sdroege: Why the signal parameter? You always call it with the same

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+{
+  if (self->priv->dispatch_to_main_context) {
+    MediaInfoUpdatedSignalData *data = g_slice_new
+                                        (MediaInfoUpdatedSignalData);
+
+    /* unref the previous media info object */
+    if (self->priv->prev_media_info)
+      g_object_unref (self->priv->prev_media_info);
+    self->priv->prev_media_info = self->priv->media_info;
+    self->priv->media_info = info;
```
sdroege: Just always keep around the very latest one

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    data->info   = self->priv->media_info;
+    data->signal = signal;
+
+    g_main_context_invoke_full (self->priv->application_context,
+                    G_PRIORITY_DEFAULT,
+                    media_info_updated_dispatch,
+                    data,
+                    (GDestroyNotify) free_media_info_updated_signal_data);
+  } else {
+    g_signal_emit (self, signals[signal], 0);
```
sdroege: In this case you're emitting the signal without the actual media info, and also you don't update self->priv->media_info

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+      if (gst_structure_get_fraction (s, "framerate", &fps_n, &fps_d)) {
+        info->framerate_num = fps_n;
+        info->framerate_denom = fps_d;
+      }
+
+      if (gst_structure_get_fraction (s, "pixel-aspect-ratio",
+            &par_n, &par_d)) {
+        info->par_num = par_n;
+        info->par_denom = par_d;
+      }
```
sdroege: By default if there is no par in the caps it should be 1/1

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+      if (gst_structure_get_int (s, "width", &width))
+        info->width = width;
+
+      if (gst_structure_get_int (s, "height", &height))
+        info->height = height;
+
+      if (gst_structure_get_fraction (s, "framerate", &fps_n, &fps_d)) {
+        info->framerate_num = fps_n;
+        info->framerate_denom = fps_d;
+      }
```
sdroege: By default if there is no framerate in the caps it should be 0/1

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1144,458 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  info->stream_list = g_list_append (info->stream_list, ret);
+}
+
+static GstPlayerMediaInfo*
+gst_player_media_info_new (GstPlayer *self)
+{
+  GstPlayerMediaInfo  *info;
+
+  info = g_object_new (GST_TYPE_PLAYER_MEDIA_INFO, NULL);
+  info->uri = g_strdup (gst_player_get_uri (self));
```
sdroege: get_uri() already gives you (or should at least) a new copy

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2154,163 @@ gst_player_get_pipeline (GstPlayer * self)
...
+/**
+ * gst_player_get_media_info:
+ * @player: #GstPlayer instance
+ *
+ * Returns: (transfer full): The internal media info instance.
+ */
+GstPlayerMediaInfo*
+gst_player_get_media_info (GstPlayer *self)
+{
+  return self->priv->media_info;
```
sdroege: Need to return a new reference here, and also needs locking (also in the other code where you actually set the media info)

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2154,163 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * Returns: (transfer full): current audio track
+ */
+GstPlayerAudioInfo*
+gst_player_get_audio_track (GstPlayer *self)
+{
+  if (!is_track_enabled (self, GST_PLAY_FLAG_AUDIO))
+    return NULL;
+
+  return (GstPlayerAudioInfo*)
+       get_current_stream (self, "current-audio", self->priv->audioinfo_list);
```
sdroege: Also here (and in the others) should be a new reference, and locking

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2154,163 @@ gst_player_get_pipeline (GstPlayer * self)
...
+}
+
+/*
+ * gst_player_audio_track_enable:
+ * @player: #GstPlayer instance
+ *
+ * Enable the current audio track
+ */
+void
+gst_player_audio_track_enable (GstPlayer *self)
```
sdroege: Maybe just one method? gst_player_audio_set_enabled(self, enabled)? And the same for video/subs

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2154,163 @@ gst_player_get_pipeline (GstPlayer * self)
...
+      get_current_stream (self, "current-text", self->priv->subtitleinfo_list);
+}
+
+/**
+ * gst_player_set_track:
+ * @player: #GstPlayer instance
+ * @s: #GstPlayerStreamInfo instance
+ */
+void
+gst_player_set_track (GstPlayer *self, GstPlayerStreamInfo *s)
```
sdroege: IMHO just gst_player_set_audio_track(self, index) (and the same for video/subs)

sdroege:
Thanks, a good step further :) I added various comments but overall I think the copy logic for the media info is unnecessarily complicated

I think it should be possible to just have a single media_info stored in GstPlayer all the time (and no other lists and stuff), and then whenever you get a signal to update something you either create a completely new one or grab the previous one and create a modified copy (with most streams just being references of the old ones)

lib/gst/player/gstplayer.c
```diff
@@ -958,6 +980,7 @@ state_changed_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
       gst_element_query_duration (self->priv->playbin, GST_FORMAT_TIME,
           &duration);
       emit_duration_changed (self, duration);
+      self->priv->is_prerolled = TRUE;
```
sdroege: The idea would be to only create the initial media_info here (for all the streams, etc) and just keep media_info at NULL before that (and if you get all the various signals when media_info==NULL you just ignore them).

codomania: Thanks for review comment, let me take a stab at easing the copy logic.

sdroege:
New version in https://github.com/sdroege/gst-player/pull/15

PR #15, opened 2015-04-20
mediainfo-v2

codomania:
I have tried to ease the copy logic. If possible give a quick review and let me know if I am on right path.

### DIFF
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -44,9 +44,11 @@
  */
 
 #include "gstplayer.h"
+#include "gstplayer-media-info-private.h"
 
 #include <gst/gst.h>
 #include <gst/video/video.h>
+#include <gst/tag/tag.h>
 
 GST_DEBUG_CATEGORY_STATIC (gst_player_debug);
 #define GST_CAT_DEFAULT gst_player_debug
@@ -85,8 +87,16 @@ enum
   SIGNAL_END_OF_STREAM,
   SIGNAL_ERROR,
   SIGNAL_VIDEO_DIMENSIONS_CHANGED,
+  SIGNAL_MEDIA_INFO_UPDATED,
   SIGNAL_LAST
 };
+
+enum {
+  GST_PLAY_FLAG_VIDEO = (1 << 0),
+  GST_PLAY_FLAG_AUDIO = (1 << 1),
+  GST_PLAY_FLAG_SUBTITLE = (1 << 2)
+};
+
 
 struct _GstPlayerPrivate
 {
@@ -111,6 +121,9 @@ struct _GstPlayerPrivate
 
   GstPlayerState app_state;
   gint buffering;
+
+  GstPlayerMediaInfo  *media_info;  /* current media info */
+  GstPlayerMediaInfo  *media_info_copy; /* last emited media info */
 
   /* Protected by lock */
   gboolean seek_pending;        /* Only set from main context */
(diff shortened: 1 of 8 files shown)


lib/gst/player/gstplayer.c
```diff
@@ -112,6 +122,9 @@ struct _GstPlayerPrivate
   GstPlayerState app_state;
   gint buffering;
 
+  GstPlayerMediaInfo  *media_info;  /* current media info */
+  GstPlayerMediaInfo  *media_info_copy; /* last emited media info */
```
sdroege: Why do you store both? Just always store the very latest one, and replace that with a modified copy after you updated it

lib/gst/player/gstplayer.c
```diff
@@ -941,6 +986,14 @@ state_changed_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
 
       GST_DEBUG_OBJECT (self, "Initial PAUSED - pre-rolled");
 
+      g_mutex_lock (&self->priv->lock);
+      self->priv->media_info = gst_player_media_info_create (self);
+      g_mutex_unlock (&self->priv->lock);
+      self->priv->media_info_copy = gst_player_media_info_copy 
+                                      (self->priv->media_info);
+      emit_media_updated_signal (self, self->priv->media_info_copy,
+                                  SIGNAL_MEDIA_INFO_UPDATED);
```
sdroege: E.g. here just do
self->priv->media_info = ...
media_info = g_object_ref (self->priv->media_info);
g_mutex_unlock();
emit(media_info);
g_object_unref(media_info);

sdroege:
Looks in the right direction, yes :) Just make sure to properly lock/unlock the mutex wherever required. Especially never emit a signal while the mutex is locked, and keep in mind that all the playbin signals can come from arbitrary threads, as can come all the function calls by the application.

lib/gst/player/gstplayer.c
codomania: Here is how code works:

- If media_info object is created then *tags-changed callback update tags, caps and info in media_info object and creates a new GstPlayerMediaInfo object by copying all the contents from media_info. Finally emit the newly created object and also update media_info_copy to points  this newly created variable.
- So next time when callback is executed the same thing happens except this time we also check if media_info_copy is not NULL, if not NULL then unref the media_info_copy and update it with newly created GstPlayerMediaInfo.

So media_info_copy points the immutable GstPlayerMediaInfo object created and emited to app, and since app does not unref the object hence I needed a way to unref the object by keeping a reference in media_info_copy.

Does it make sense ?

sdroege: Why not:
- whenever something changes, create a new one or copy. Then modify that one
- if had an old one, unref that one
- store the new one
- emit the signal with the new one

It's not needed to keep the old one around until you emitted the signal for the new one

codomania: I guess one part which I am still not clear is what will happen in this scenario:

1. we emited a GstPlayerMediaInfo object (media_info) 
2. something changes and we created a new one.
3. now I have a old one from step#1, so lets unref this one. (But old one is still being referred by app and if it tries to access members of old media_info before receiving new one may cause segfault).
4. store the new one
5. emit the signal with new one.

So to solve this I was unrefing the old one only after we successfully emited the new one. I am doing something like this:

1. whenever something changes, create a new copy.  
info = gst_player_media_info_copy (self->priv->media_info);
2. emit new one
emit_media_info_updated_signal (info)
3. if had an old one, unref that one
if (self->priv->media_info_copy)
 g_object_unref (self->priv->media_info_copy)
4. point old one to new one
self->priv->media_info_copy = info.

It possible that I am over thinking and making it unnecessary complicated. Thanks for your patience and detail review feedbacks.

sdroege: The app will get the media info in some callback like this

    static void media_info_updated_cb (GstPlayer * player, GstPlayerMediaInfo *info, gpointer user_data);

The app can use the info while inside the scope of that function, but not afterwards anymore. If it wants to store the info elsewhere, it must get another reference with g_object_ref() and later unref() it again.

codomania: Okay that clarifies it. I was expecting that info maybe used outside the scope of app callback.

sdroege: If it is, then the application has to take another reference. Sorry that I didn't notice before that this was what caused the confusion :)

sdroege:
New version in https://github.com/sdroege/gst-player/pull/17

PR #16, opened 2015-04-20
Changing playback rate

rishabh7m:


sdroege:
Thanks for these changes, but please squash these separate commits into one. There should be one commit per functional change in the end, not many work-in-progress commits.

lib/gst/player/gstplayer.c
```diff
@@ -688,6 +689,44 @@ eos_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  
+  gst_element_set_state (self->priv->playbin, GST_STATE_PLAYING);
+
+  return G_SOURCE_REMOVE;  
+}
+
+void gst_player_set_playback_rate (GstPlayer * self, gdouble rate)
+{
+  g_return_if_fail (GST_IS_PLAYER (self));
+  self->priv->rate = rate;
```
sdroege: Normal seek events also have to take the rate into account. I think it would make sense to just update the rate, and the reuse the existing seeking logic (plus adaptions) instead of writing another completely separate seek function.

sdroege:
Replaced by https://github.com/sdroege/gst-player/pull/24

PR #17, opened 2015-04-21
mediainfo-v3

codomania:
Your review feedback is addressed in this patch set.

### DIFF
--- a/lib/gst/player/gstplayer-media-info.c
+++ b/lib/gst/player/gstplayer-media-info.c
@@ -1,0 +1,649 @@
+#include "gstplayer-media-info.h"
+#include "gstplayer-media-info-private.h"
+
+/* Per-stream information */
+G_DEFINE_TYPE (GstPlayerStreamInfo, gst_player_stream_info, G_TYPE_OBJECT);
+
+static void
+gst_player_stream_info_init (GstPlayerStreamInfo * sinfo)
+{
+  sinfo->stream_index = -1;
+}
+
+static void
+gst_player_stream_info_finalize (GObject * object)
+{
+  GstPlayerStreamInfo *sinfo = GST_PLAYER_STREAM_INFO (object);
+
+  if (sinfo->caps)
+    gst_caps_unref (sinfo->caps);
+
+  if (sinfo->tags)
+    gst_tag_list_unref (sinfo->tags);
+
+  G_OBJECT_CLASS (gst_player_stream_info_parent_class)->finalize (object);
+}
+
+static void
+gst_player_stream_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_stream_info_finalize;
+}
+
+/**
+ * gst_player_stream_info_get_stream_index:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: the stream index of this stream.
+ */
+gint
+gst_player_stream_info_get_stream_index (const GstPlayerStreamInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_STREAM_INFO (info));
+
+  return info->stream_index;
+}
+
+/**
+ * gst_player_stream_info_get_stream_type_nick:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: a human readable name for the stream type of the given @info (ex : "audio",
+ * "video",...).
+ */
+const gchar*
+gst_player_stream_info_get_stream_type_nick (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  if (GST_IS_PLAYER_VIDEO_INFO (info))
+    return "video";
+
+  if (GST_IS_PLAYER_AUDIO_INFO (info))
+    return "audio";
+
+  if (GST_IS_PLAYER_SUBTITLE_INFO (info))
+    return "subtitle";
+
+  return NULL;
+}
+
+/**
+ * gst_player_stream_info_get_tags:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: (transfer none): the tags contained in this stream.
+ */
+GstTagList*
+gst_player_stream_info_get_stream_tags (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  return info->tags;
+}
+
+/**
+ * gst_player_stream_info_get_caps:
+ * @info: a #GstPlayerStreamInfo
+ *
+ * Returns: (transfer none): the #GstCaps of the stream.
+ */
+GstCaps*
+gst_player_stream_info_get_stream_caps (const GstPlayerStreamInfo *info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_STREAM_INFO (info), NULL);
+
+  return info->caps;
+}
+
+/* Video information */
+G_DEFINE_TYPE (GstPlayerVideoInfo, gst_player_video_info,
+    GST_TYPE_PLAYER_STREAM_INFO);
+
+static void
+gst_player_video_info_init (GstPlayerVideoInfo * info)
+{
+  info->width = -1;
+  info->height = -1;
+  info->framerate_num = 0;
+  info->framerate_denom = 1;
+  info->par_num = 1;
+  info->par_denom = 1;
+}
+
+static void
+gst_player_video_info_class_init (GObjectClass * klass)
+{
+  /* nothing to do here */
+}
+
+/**
+ * gst_player_video_info_get_width:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the width of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_width (const GstPlayerVideoInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
+
+  return info->width;
+}
+
+/**
+ * gst_player_video_info_get_height:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the height of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_height (const GstPlayerVideoInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
+
+  return info->height;
+}
+
+/**
+ * gst_player_video_info_get_framerate_num:
+ * @info: a #GstPlayerVideoInfo
+ *
+ */
+void
+gst_player_video_info_get_framerate (const GstPlayerVideoInfo *info,
+  gint *fps_n, gint *fps_d)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
+
+  *fps_n = info->framerate_num;
+  *fps_d = info->framerate_denom;
+}
+
+/**
+ * gst_player_video_info_get_pixel_aspect_ratio:
+ * @info: a #GstPlayerVideoInfo
+ *
+ */
+void
+gst_player_video_info_get_pixel_aspect_ratio (const GstPlayerVideoInfo *info,
+  guint *par_d, guint *par_n)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
+
+  *par_n = info->par_num;
+  *par_d = info->par_denom;
+}
+
+/**
+ * gst_player_video_info_get_bitrate:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the current bitrate of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_bitrate (const GstPlayerVideoInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
+
+  return info->bitrate;
+}
+
+/**
+ * gst_player_video_info_get_max_bitrate:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the maximum bitrate of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_max_bitrate (const GstPlayerVideoInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
+
+  return info->max_bitrate;
+}
+
+/* Audio information */
+G_DEFINE_TYPE (GstPlayerAudioInfo, gst_player_audio_info,
+    GST_TYPE_PLAYER_STREAM_INFO);
+
+static void
+gst_player_audio_info_init (GstPlayerAudioInfo * info)
+{
+  /* nothing to do here */
+}
+
+static void
+gst_player_audio_info_finalize (GObject * object)
+{
+  GstPlayerAudioInfo *info = GST_PLAYER_AUDIO_INFO (object);
+
+  g_free (info->language);
+
+  G_OBJECT_CLASS (gst_player_audio_info_parent_class)->finalize (object);
+}
+
+static void
+gst_player_audio_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_audio_info_finalize;
+}
+
+/**
+ * gst_player_audio_info_get_language:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the language of the stream, or NULL if unknown.
+ */
+const gchar*
+gst_player_audio_info_get_language(const GstPlayerAudioInfo* info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_AUDIO_INFO (info), NULL);
+
+ return info->language;
+}
+
+/**
+ * gst_player_audio_info_get_channels:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the number of audio channels in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_channels (const GstPlayerAudioInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
+
+  return info->channels;
+}
+
+/**
+ * gst_player_audio_info_get_sample_rate:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the audio sample rate in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_sample_rate (const GstPlayerAudioInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
+
+  return info->sample_rate;
+}
+
+/**
+ * gst_player_audio_info_get_bitrate:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the audio bitrate in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_bitrate (const GstPlayerAudioInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
+
+  return info->bitrate;
+}
+
+/**
+ * gst_player_audio_info_get_max_bitrate:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the audio maximum bitrate in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_max_bitrate (const GstPlayerAudioInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
+
+  return info->max_bitrate;
+}
+
+/* Subtitle information */
+G_DEFINE_TYPE (GstPlayerSubtitleInfo, gst_player_subtitle_info,
+    GST_TYPE_PLAYER_STREAM_INFO);
+
+static void
+gst_player_subtitle_info_init (GstPlayerSubtitleInfo * info)
+{
+  /* nothing to do */
+}
+
+static void
+gst_player_subtitle_info_finalize (GObject * object)
+{
+  GstPlayerSubtitleInfo *info = GST_PLAYER_SUBTITLE_INFO (object);
+
+  g_free (info->language);
+
+  G_OBJECT_CLASS (gst_player_subtitle_info_parent_class)->finalize (object);
+}
+
+static void
+gst_player_subtitle_info_class_init (GObjectClass * klass)
+{
+  klass->finalize = gst_player_subtitle_info_finalize;
+}
+
+/**
+ * gst_player_subtitle_get_language:
+ * @info: a #GstPlayerSubtitleInfo
+ *
+ * Returns: the language of the stream, or NULL if unknown.
+ */
+const gchar*
+gst_player_subtitle_get_language( const GstPlayerSubtitleInfo* info)
+{
+  g_return_val_if_fail (GST_IS_PLAYER_SUBTITLE_INFO (info), NULL);
+
+  return info->language;
+}
+
+/* Global media information */
+G_DEFINE_TYPE (GstPlayerMediaInfo, gst_player_media_info, G_TYPE_OBJECT);
+
+static void
+gst_player_media_info_init (GstPlayerMediaInfo * info)
+{
+  info->duration = -1;
+}
+
+static void
+gst_player_media_info_finalize (GObject * object)
+{
+  GstPlayerMediaInfo  *info = GST_PLAYER_MEDIA_INFO (object);
+
+  g_free (info->uri);
+
+  if (info->stream_list)
+    g_list_free_full (info->stream_list, g_object_unref);
+
+  G_OBJECT_CLASS (gst_player_media_info_parent_class)->finalize (object);
+}
+
+static void
+gst_player_media_info_class_init (GstPlayerMediaInfoClass * klass)
+{
+  GObjectClass  *oclass = (GObjectClass *) klass;
+
+  oclass->finalize = gst_player_media_info_finalize;
+}
+
+static GstPlayerVideoInfo*
+gst_player_video_info_new (void)
+{
+  return g_object_new (GST_TYPE_PLAYER_VIDEO_INFO, NULL);
+}
+
+static GstPlayerAudioInfo*
+gst_player_audio_info_new (void)
+{
+  return g_object_new (GST_TYPE_PLAYER_AUDIO_INFO, NULL);
+}
+
+static GstPlayerSubtitleInfo*
+gst_player_subtitle_info_new (void)
+{
+  return g_object_new (GST_TYPE_PLAYER_SUBTITLE_INFO, NULL);
+}
+
+static GstPlayerStreamInfo*
+gst_player_video_info_copy (GstPlayerVideoInfo *ref)
+{
+  GstPlayerVideoInfo *ret;
+
+  ret = gst_player_video_info_new ();
+
+  ret->width = ref->width;
+  ret->height  = ref->height;
+  ret->framerate_num = ref->framerate_num;
+  ret->framerate_denom = ref->framerate_denom;
+  ret->par_num = ref->par_num;
+  ret->par_denom = ref->par_denom;
+  ret->bitrate = ref->bitrate;
+  ret->max_bitrate = ref->max_bitrate;
+
+  return (GstPlayerStreamInfo*) ret;
+}
+
+static GstPlayerStreamInfo*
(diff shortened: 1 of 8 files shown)


codomania:
changes since v2:

* remove media_info_copy. now will have only one copy of media_info in GstPlayer instance.
* return immutable object from the following function, this is mainly because internal media_info object content will change while app using the instance.
  * gst_player_get_media_info ()
  * gst_player_get_audio_track, gst_player_get_video_track and gst_player_get_subtitle_track ()
* updated gst-play.c to use gst_player_get_media_info() to obtain the media_info instance

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_video_info_get_width:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the width of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_width (const GstPlayerVideoInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
```
sdroege: g_return_val_if_fail() whenever your function has a non-void return type :)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_video_info_get_height:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the height of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_height (const GstPlayerVideoInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
```
sdroege: g_return_val_if_fail()

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_video_info_get_bitrate:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the current bitrate of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_bitrate (const GstPlayerVideoInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
```
sdroege: here too (maybe initialize the bitrates as -1 if unknown too)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_video_info_get_max_bitrate:
+ * @info: a #GstPlayerVideoInfo
+ *
+ * Returns: the maximum bitrate of video in #GstPlayerVideoInfo.
+ */
+gint
+gst_player_video_info_get_max_bitrate (const GstPlayerVideoInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_VIDEO_INFO (info));
```
sdroege: here  too

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_audio_info_get_channels:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the number of audio channels in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_channels (const GstPlayerAudioInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
```
sdroege: here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_audio_info_get_sample_rate:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the audio sample rate in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_sample_rate (const GstPlayerAudioInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
```
sdroege: here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_audio_info_get_bitrate:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the audio bitrate in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_bitrate (const GstPlayerAudioInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
```
sdroege: here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_audio_info_get_max_bitrate:
+ * @info: a #GstPlayerAudioInfo
+ *
+ * Returns: the audio maximum bitrate in #GstPlayerAudioInfo.
+ */
+gint
+gst_player_audio_info_get_max_bitrate (const GstPlayerAudioInfo* info)
+{
+  g_return_if_fail (GST_IS_PLAYER_AUDIO_INFO (info));
```
sdroege: here

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+  ret = gst_player_video_info_new ();
+
+  ret->width = ref->width;
+  ret->height  = ref->height;
+  ret->framerate_num = ref->framerate_num;
+  ret->framerate_denom = ref->framerate_denom;
+  ret->par_num = ref->par_num;
+  ret->par_denom = ref->par_denom;
+  ret->bitrate = ref->bitrate;
+  ret->max_bitrate = ref->max_bitrate;
```
sdroege: What about copying the caps, taglist, etc from GstPlayerStreamInfo?

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+
+GstPlayerStreamInfo*
+gst_player_stream_info_copy (GstPlayerStreamInfo *ref)
+{
+  GstPlayerStreamInfo *info = NULL;
+
+  if (GST_IS_PLAYER_VIDEO_INFO(ref))
+    info = gst_player_video_info_copy ((GstPlayerVideoInfo*)ref);
+
+  if (GST_IS_PLAYER_AUDIO_INFO(ref))
```
sdroege: else if

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+    info = gst_player_audio_info_copy ((GstPlayerAudioInfo*)ref);
+
+  if (GST_IS_PLAYER_SUBTITLE_INFO(ref))
```
sdroege: else if

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+  GstPlayerMediaInfo *info;
+
+  if (!ref)
+    return NULL;
+
+  info = gst_player_media_info_new (ref->uri);
+
+  for (l = ref->stream_list; l != NULL; l = l->next)
+    info->stream_list = g_list_append (info->stream_list,
+                  gst_player_stream_info_copy((GstPlayerStreamInfo*)l->data));
```
sdroege: What about copying the duration and other things from GstPlayerMediaInfo?

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+
+GstPlayerStreamInfo*
+gst_player_stream_info_new (gint stream_index, GType type)
+{
+  GstPlayerStreamInfo *info = NULL;
+
+  if (type == GST_TYPE_PLAYER_AUDIO_INFO)
+    info = (GstPlayerStreamInfo*) gst_player_audio_info_new ();
+
+  if (type == GST_TYPE_PLAYER_VIDEO_INFO)
```
sdroege: else if

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+    info = (GstPlayerStreamInfo*) gst_player_video_info_new ();
+
+  if (type == GST_TYPE_PLAYER_SUBTITLE_INFO)
```
sdroege: else if

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+}
+
+GstPlayerMediaInfo*
+gst_player_media_info_new (const gchar *uri)
+{
+  GstPlayerMediaInfo *info;
+
+  info = g_object_new (GST_TYPE_PLAYER_MEDIA_INFO, NULL);
+  if (uri)
+    info->uri = g_strdup (uri);
```
sdroege: I think this must be called with a valid URI, add a g_return_val_if_fail(uri != NULL, NULL) at the top :)

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+  g_return_val_if_fail (GST_IS_PLAYER_MEDIA_INFO (info), NULL);
+
+  return info->stream_list;
+}
+
+/**
+ * gst_player_get_video_streams:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: (transfer full) (element-type GstPlayerVideoInfo): A #GList of
```
sdroege: I don't like that get_stream_list() is (transfer full), but the media specific ones are not. Maybe make the media specific ones (transfer none) too by caching the media specific lists inside GstPlayerMediaInfo

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+}
+
+/**
+ * gst_player_stream_info_list_free:
+ * @list: (element-type GstPlayerStreamInfo): A #GList of #GstPlayerStreamInfo
+ *
+ * frees the #GList.
+ */
+void
+gst_player_stream_info_list_free (GList *list)
```
sdroege: ... then this could also go away

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,649 @@
...
+/**
+ * gst_player_media_info_get_duration:
+ * @info: a #GstPlayerMediaInfo
+ *
+ * Returns: duration of the media.
+ */
+GstClockTime
+gst_player_media_info_get_duration (const GstPlayerMediaInfo *info)
+{
+  g_return_if_fail (GST_IS_PLAYER_MEDIA_INFO (info));
```
sdroege: g_return_val_if_fail()

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    g_mutex_unlock (&self->priv->lock);
+    data->signal = signal;
+
+    g_main_context_invoke_full (self->priv->application_context,
+                    G_PRIORITY_DEFAULT,
+                    media_info_updated_dispatch,
+                    data,
+                    (GDestroyNotify) free_media_info_updated_signal_data);
+  } else {
+    g_signal_emit (self, signals[signal], 0);
```
sdroege: You don't provide the media-info here, and it should be
lock()
media_info = g_object_ref (priv->media_info);
unlock();
emit(media_info);
unref(media_info);

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  if (self->priv->dispatch_to_main_context) {
+    MediaInfoUpdatedSignalData *data = g_slice_new
+                                        (MediaInfoUpdatedSignalData);
+
+    data->player = self;
+    g_mutex_lock (&self->priv->lock);
+    if (self->priv->media_info)
+      g_object_unref (self->priv->media_info);
+    self->priv->media_info = info;
+    data->info   = info;
```
sdroege: You need to store a new ref in data, and then unref it later

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    g_mutex_unlock (&self->priv->lock);
+    data->signal = signal;
```
sdroege: The signal is always the same, so you don't really need that

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+static void
+free_media_info_updated_signal_data (MediaInfoUpdatedSignalData *data)
+{
+  g_slice_free (MediaInfoUpdatedSignalData, data);
+}
+
+static void
+emit_media_updated_signal (GstPlayer *self, GstPlayerMediaInfo *info,
+  guint signal)
```
sdroege: The signal is always the same, so you don't really need that

lib/gst/player/gstplayer.c
sdroege: Later means: in the free function for the data

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  if (type == GST_TYPE_PLAYER_AUDIO_INFO)
+    g_signal_emit_by_name (G_OBJECT (self->priv->playbin),
+                          "get-audio-pad", stream_index, &pad);
+
+  if (type == GST_TYPE_PLAYER_SUBTITLE_INFO)
+    g_signal_emit_by_name (G_OBJECT (self->priv->playbin),
+                          "get-text-pad", stream_index, &pad);
+
+  if (pad)
+    caps = gst_pad_get_current_caps (pad);
```
sdroege: You need to unref the pad afterwards

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+      if (gst_structure_get_int (s, "width", &width))
+        info->width = width;
+
+      if (gst_structure_get_int (s, "height", &height))
+        info->height = height;
+
+      if (gst_structure_get_fraction (s, "framerate", &fps_n, &fps_d)) {
+        info->framerate_num = fps_n;
+        info->framerate_denom = fps_d;
```
sdroege: else: 0/1

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+      }
+
+      if (gst_structure_get_fraction (s, "pixel-aspect-ratio",
+            &par_n, &par_d)) {
+        info->par_num = par_n;
+        info->par_denom = par_d;
```
sdroege: else: 1/1

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    if (s) {
+      gint width, height;
+      gint fps_n, fps_d;
+      gint par_n, par_d;
+
+      if (gst_structure_get_int (s, "width", &width))
+        info->width = width;
+
+      if (gst_structure_get_int (s, "height", &height))
+        info->height = height;
```
sdroege: For width/height: else 0 (or -1)

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    gchar *lang_code = NULL;
+
+    if (gst_tag_list_get_string (stream_info->tags,
+          GST_TAG_LANGUAGE_CODE, &lang_code)) {
+      info->language = g_strdup (gst_tag_get_language_name (lang_code));
+      g_free (lang_code);
+    }
+    else {
+      gst_tag_list_get_string (stream_info->tags,
+          GST_TAG_LANGUAGE_CODE, &info->language);
```
sdroege: Shouldn't this just set info->language to NULL (and free it)?

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    guint bitrate, max_bitrate;
+
+    if (gst_tag_list_get_uint (stream_info->tags, GST_TAG_BITRATE,
+          &bitrate) || gst_tag_list_get_uint (stream_info->tags,
+            GST_TAG_NOMINAL_BITRATE, &bitrate))
+      info->bitrate = bitrate;
+
+    if (gst_tag_list_get_uint (stream_info->tags,
+          GST_TAG_MAXIMUM_BITRATE, &max_bitrate))
+      info->max_bitrate = max_bitrate;
```
sdroege: Reset if not in the tags

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+    s = gst_caps_get_structure (stream_info->caps, 0);
+    if (s) {
+      gint rate, channels;
+
+      if (gst_structure_get_int (s, "rate", &rate))
+        info->sample_rate = rate;
+
+      if (gst_structure_get_int (s, "channels", &channels))
+        info->channels = channels;
```
sdroege: Also here reset if not in the caps

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+          GST_TAG_MAXIMUM_BITRATE, &max_bitrate))
+      info->max_bitrate = max_bitrate;
+
+    if (gst_tag_list_get_string (stream_info->tags,
+          GST_TAG_LANGUAGE_CODE, &lang_code)) {
+      info->language = g_strdup (gst_tag_get_language_name (lang_code));
+      g_free (lang_code);
+    }
+    else {
+      gst_tag_list_get_string (stream_info->tags,
```
sdroege: And some comments here about bitrate and language

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+  for (i = 0; i < total; i++) {
+    s = gst_player_stream_info_find (self, media_info, type, i);
+    if (!s) {
+      s = gst_player_stream_info_new (i, type);
+      media_info->stream_list = g_list_append (media_info->stream_list, s);
+      GST_DEBUG_OBJECT (self, "create %s stream stream_index: %d",
+        gst_player_stream_info_get_stream_type_nick(s), i);
+    }
+    gst_player_stream_info_update_tags (self, s);
```
sdroege: This is more an update_tags_and_caps() :)

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  g_mutex_unlock (&self->priv->lock);
+}
+
+static void
+subtitle_changed_cb (GObject *object, gpointer user_data)
+{
+  GstPlayer *self = GST_PLAYER(user_data);
+
+  g_mutex_lock (&self->priv->lock);
+  gst_player_streams_info_create (self, self->priv->media_info,
```
sdroege: Here (and in the other callbacks) you're modifying existing stream infos but we said the mediainfo/streaminfos are immutable

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2181,200 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ *
+ * the caller should free it with g_object_unref()
+ */
+GstPlayerMediaInfo*
+gst_player_get_media_info (GstPlayer *self)
+{
+  GstPlayerMediaInfo *info;
+
+  g_mutex_lock (&self->priv->lock);
+  info = gst_player_media_info_copy (self->priv->media_info);
```
sdroege: No need for a copy, can just be a new reference. That's why we wanted it to be immutable among other things :)

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1167,464 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+{
+  gint current;
+  GstPlayerStreamInfo *info;
+
+  g_object_get (G_OBJECT(self->priv->playbin), prop, &current, NULL);
+  g_mutex_lock (&self->priv->lock);
+  info = gst_player_stream_info_find (self, self->priv->media_info,
+            type, current);
+  if (info)
+    info = gst_player_stream_info_copy (info);
```
sdroege: A new reference seems enough here, you're never going to change them

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2181,200 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * gst_player_set_audio_track:
+ * @player: #GstPlayer instance
+ * @index: stream index
+ */
+gboolean
+gst_player_set_audio_track (GstPlayer *self, gint stream_index)
+{
+  GstPlayerStreamInfo *info;
+
+  info = gst_player_stream_info_find (self, self->priv->media_info,
```
sdroege: You have to unref this afterwards

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2181,200 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * gst_player_set_video_track:
+ * @player: #GstPlayer instance
+ * @index: stream index
+ */
+gboolean
+gst_player_set_video_track (GstPlayer *self, gint stream_index)
+{
+  GstPlayerStreamInfo *info;
+
+  info = gst_player_stream_info_find (self, self->priv->media_info,
```
sdroege: and here

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2181,200 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * gst_player_set_subtitle_track:
+ * @player: #GstPlayer instance
+ * @index: stream index
+ */
+gboolean
+gst_player_set_subtitle_track (GstPlayer *self, gint stream_index)
+{
+  GstPlayerStreamInfo *info;
+
+  info = gst_player_stream_info_find (self, self->priv->media_info,
```
sdroege: and here

lib/gst/player/gstplayer.h
```diff
@@ -109,6 +110,32 @@ void         gst_player_set_window_handle             (GstPlayer    * player,
...
+
+gboolean      gst_player_set_video_track              (GstPlayer    *player,
+                                                       gint stream_index);
+
+gboolean      gst_player_set_subtitle_track           (GstPlayer    *player,
+                                                       gint stream_index);
+
+GstPlayerMediaInfo * gst_player_get_media_info        (GstPlayer    * player);
+
+GstPlayerAudioInfo * gst_player_get_audio_track       (GstPlayer    * player);
```
sdroege: get_current_*_track() IMHO

sdroege:
Thanks, almost there :) As you see I commented on a lot of little things now, the difficult parts are almost all done now

lib/gst/player/gstplayer.c
codomania: Okay maybe I am missing something or have some confusion.

* self->priv->media_info is pointing to internal media_info. this media_info is constantly been updated by *-tags-changed or *-changed callbacks.
* when *-tags-changed or *-changed callback is executed then we update the content of self->priv->media_info and create a new copy from the current media_info and emit the newly created copy to user. In this way, the media_info received by user app is immutable
so our sequence is:
lock()
new_media_info = gst_player_media_info_copy (self->priv->media_info);
unlock()
emit (new_media_info)
unref(new_meda_info)

Do you see anything wrong in this approach ?

If we give reference of self->priv->media_info then its not immutable and it can change while user app is still inside the emit signal callback.

sdroege: Ah now I understand what you're doing there! That also makes sense, but it should be documented in comments for the private media info (maybe in GstPlayerPrivate struct definition).

I would do the copying here in emit_signal() though, and then always when the user calls one of the getter APIs. To keep the copy at exactly the place where it is needed, and otherwise just work on our mutable internal version.

codomania:
Thanks. Yes agree at least now skeleton is in place and we should be able to change it to meet our goal. Looks like I still have some confusion in my immutable understanding vs yours :(
 
Our current implementation is like this: 
* We discard all the *-tags-changed and *-changed callback before pipeline preroll. Once we detect the pipeline is prerolled then we:
  * create a new media_info object and store it in self->priv->media_info.   
* After preroll, when get *-tags-changed or *-changed callback then we update the content of self->priv->media_info and after updates create a new copy from self->priv->media_info and emit the newly created copy to user. In this way, the object received by user app is immutable and also self->priv->media_info has all the latest media information.
so our sequence is: <br>
lock() <br>
new_media_info = gst_player_media_info_copy (self->priv->media_info); <br>
unlock() <br>
emit (new_media_info)<br>
unref(new_meda_info)<br>

If we give reference of self->priv->media_info in emit signal then object will not be immutable and it can changed while user app is still inside the emit signal callback.

This same logic holds true for gst_player_get_media_info(), gst_player_current_*_track().  I was thinking that whatever object we sent to user should never be changed after it given to user app.

lib/gst/player/gstplayer.c
codomania: great. I will make those suggested changes and also add some verbatim regarding this in our header file and function.

lib/gst/player/gstplayer-media-info.c
codomania: sure sound like good idea. I will create GList *_cachelist in GstPlayerMediaInfo which will point to audio, video and subtitle list and then we can simply return that *_cachelist from here.

lib/gst/player/gstplayer-media-info.c
codomania: good catch, I will add it.

lib/gst/player/gstplayer.c
codomania: Actually its getting unref'ed in our stream finalize method. The caps returned from this function is saved in (GstPlayerStreamInfo)->caps and as part of our finalize method we call <br>
if (caps) <br> gst_caps_unref(info->caps) <br>

lib/gst/player/gstplayer.c
codomania: one question on this section, since we are initializing <br>
width=-1, height=-1, fps_n=0, fps_d=1 and par_n=par_d=1<br>
in our class_init section hence else part is not required. If() fails then we are not setting that field hence the field contains the default.

lib/gst/player/gstplayer.c
codomania: agreed :)

lib/gst/player/gstplayer.c
sdroege: The pad, not the caps :) The caps are fine

lib/gst/player/gstplayer-media-info.c
sdroege: If you want to prevent thread-safety headaches here, always create those lists together with the GstPlayerMediaInfo

lib/gst/player/gstplayer.c
sdroege: But you might've set it to a different value before already, and now just update the stream info for new caps

codomania: never mind, while typing this i did realized one case which will warrant us to write else part and default the values. e.g if previous tag has different width/height etc.
I will simply reset the fields to default,

sdroege:
Yes, we have the same understanding here. What you say is exactly what I mean :)
I just mean that we should do the copying directly in the place where the info is needed, i.e. inside emit_media_info_changed_signal() and in the getters for the streams.

lib/gst/player/gstplayer.c
codomania: damn i suck :)

sdroege: If you did then you wouldn't have written a working patch at all and I would've just said "this is horrible" :P Don't worry, I'm sure we find reference leaks in my code too if we look close enough :)

PR #19, opened 2015-04-21
mediainfo-v5

codomania:
Changes since v3:

* g_return_val_if_fail() whenever function has a non-void return type
* initialize the bitrate, max_bitrate to -1.
* use else if(...) statement when checking stream info types.
* copy the missing duration field in gst_player_media_info_copy (ref).
* add uri NULL assertion in gst_player_media_info_new (uri).
* create three Glists to cache the stream type.
* make media specific get_stream_list to (transfer none). The function will simply return the cached stream type list.
* remove the stream_info_free_list (). We no longer need it after making media specific get_stream_list to (transfer none).
* unref the temporary GstPad in get_caps()
* simplify the emit_media_info_updated_signal ()
  * create the new media info inside the function
  * send this newly created media_info object to user app.
  * unref the newly created media_info object in signal finalize method.
* rename gst_player_get_*_track to gst_player_get_current_*_track()

### DIFF
--- a/gst-play/gst-play.c
+++ b/gst-play/gst-play.c
@@ -110,6 +110,222 @@ buffering_cb (GstPlayer * player, gint percent, GstPlay * play)
   g_print ("Buffering: %d\n", percent);
 }
 
+static void
+print_one_tag (const GstTagList *list, const gchar *tag, gpointer user_data)
+{
+  gint i, num;
+
+  num = gst_tag_list_get_tag_size (list, tag);
+  for (i = 0; i < num; ++i) {
+    const GValue *val;
+
+    val = gst_tag_list_get_value_index (list, tag, i);
+    if (G_VALUE_HOLDS_STRING (val)) {
+      g_print ("    %s : %s \n", tag, g_value_get_string (val));
+    }
+    else if (G_VALUE_HOLDS_UINT (val)) {
+      g_print ("    %s : %u \n", tag, g_value_get_uint (val));
+    }
+    else if (G_VALUE_HOLDS_DOUBLE (val)) {
+      g_print ("    %s : %g \n", tag, g_value_get_double (val));
+    }
+    else if (G_VALUE_HOLDS_BOOLEAN (val)) {
+      g_print ("    %s : %s \n", tag,
+                g_value_get_boolean (val) ? "true" : "false");
+    }
+    else if (GST_VALUE_HOLDS_DATE_TIME (val)) {
+      GstDateTime *dt = g_value_get_boxed (val);
+      gchar *dt_str = gst_date_time_to_iso8601_string (dt);
+
+      g_print ("    %s : %s \n", tag, dt_str);
+      g_free (dt_str);
+    }
+    else {
+      g_print ("    %s : tag of type '%s' \n", tag, G_VALUE_TYPE_NAME (val));
+    }
+  }
+}
+
+static void
+print_video_info (GstPlayerVideoInfo *info)
+{
+  gint fps_n, fps_d;
+  guint par_n, par_d;
+
+  if (info == NULL)
+    return;
+
+  g_print ("  width : %d\n", gst_player_video_info_get_width(info));
+  g_print ("  height : %d\n", gst_player_video_info_get_height (info));
+  g_print ("  max_bitrate : %d\n", gst_player_video_info_get_max_bitrate (info));
+  g_print ("  bitrate : %d\n", gst_player_video_info_get_bitrate (info));
+  gst_player_video_info_get_framerate (info, &fps_n, &fps_d);
+  g_print ("  frameate : %.2f\n", (gdouble) fps_n/fps_d);
+  gst_player_video_info_get_pixel_aspect_ratio (info, &par_n, &par_d);
+  g_print ("  pixel-aspect-ratio  %u:%u\n", par_n, par_d);
+}
+
+static void
+print_audio_info (GstPlayerAudioInfo *info)
+{
+  if (info == NULL)
+    return;
+
+  g_print ("  sample rate : %d\n", gst_player_audio_info_get_sample_rate (info));
+  g_print ("  channels : %d\n", gst_player_audio_info_get_channels (info));
+  g_print ("  max_bitrate : %d\n", gst_player_audio_info_get_max_bitrate (info));
+  g_print ("  bitrate : %d\n", gst_player_audio_info_get_bitrate (info));
+  g_print ("  language : %s\n", gst_player_audio_info_get_language (info));
+}
+
+static void
+print_subtitle_info (GstPlayerSubtitleInfo *info)
+{
+  if (info == NULL)
+    return;
+
+  g_print ("  language : %s\n", gst_player_subtitle_get_language (info));
+}
+
+static void
+print_all_stream_info (GstPlayerMediaInfo *media_info)
+{
+  guint count = 0;
+  GList *l, *list;
+
+  list = gst_player_media_info_get_stream_list (media_info);
+  g_print ("URI : %s\n", gst_player_media_info_get_uri (media_info));
+  g_print ("Duration: %" GST_TIME_FORMAT "\n",
+      GST_TIME_ARGS(gst_player_media_info_get_duration (media_info)));
+  for (l = list; l != NULL; l = l->next) {
+    GstTagList  *tags = NULL;
+    GstPlayerStreamInfo *stream = (GstPlayerStreamInfo*) l->data;
+
+    g_print (" Stream # %u \n", count++);
+    g_print ("  type : %s_%u\n",
+              gst_player_stream_info_get_stream_type_nick (stream),
+              gst_player_stream_info_get_stream_index (stream));
+    tags = gst_player_stream_info_get_stream_tags (stream);
+    g_print ("  taglist : \n");
+    if (tags) {
+      gst_tag_list_foreach (tags, print_one_tag, NULL);
+    }
+
+    if (GST_IS_PLAYER_VIDEO_INFO (stream))
+      print_video_info ((GstPlayerVideoInfo*)stream);
+    else if (GST_IS_PLAYER_AUDIO_INFO (stream))
+      print_audio_info ((GstPlayerAudioInfo*)stream);
+    else
+      print_subtitle_info ((GstPlayerSubtitleInfo*)stream);
+  }
+}
+
+static void
+print_all_video_stream (GstPlayerMediaInfo *media_info)
+{
+  GList *list = NULL, *l;
+
+  list = gst_player_get_video_streams (media_info);
+  if (!list)
+    return;
+
+  g_print ("All video streams\n");
+  for (l = list; l != NULL; l = l->next) {
+    GstPlayerVideoInfo  *info = (GstPlayerVideoInfo*) l->data;
+    GstPlayerStreamInfo *sinfo = (GstPlayerStreamInfo*) info;
+    g_print (" %s_%d #\n", gst_player_stream_info_get_stream_type_nick (sinfo),
+            gst_player_stream_info_get_stream_index (sinfo));
+    print_video_info (info);
+  }
+}
+
+static void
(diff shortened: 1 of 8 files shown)


codomania:
I tried the audio.ogg file tests/media directory and ran into segfault. In root cause I found that
gst_tag_list_get_string () returns 'true' even value is not set hence our logic was broken. Now I have simplified it in above commit.

gst-play/gst-play.c
```diff
@@ -110,6 +110,222 @@ buffering_cb (GstPlayer * player, gint percent, GstPlay * play)
...
+  g_print ("\n");
+  print_all_audio_stream (media_info);
+  g_print ("\n");
+  print_all_subtitle_stream (media_info);
+}
+
+static void
+media_info_cb (GstPlayer *player, GstPlayerMediaInfo *info, GstPlay *play)
+{
+}
```
sdroege: Just remove that then, or maybe print something that the media infos changed

gst-play/gst-play.c
```diff
@@ -347,6 +566,15 @@ keyboard_cb (const gchar * key_input, gpointer user_data)
   GstPlay *play = (GstPlay *) user_data;
 
   switch (g_ascii_tolower (key_input[0])) {
+    case 'i':
+      {
+        GstPlayerMediaInfo *media_info = gst_player_get_media_info (play->player);
+        print_media_info (media_info);
```
sdroege: Only calls this if media_info!=NULL, then you only need that check in one place instead of two

lib/gst/player/gstplayer-media-info.c
```diff
@@ -0,0 +1,623 @@
...
+  info->duration = ref->duration;
+
+  for (l = ref->stream_list; l != NULL; l = l->next) {
+    GstPlayerStreamInfo *s;
+
+    s = gst_player_stream_info_copy((GstPlayerStreamInfo*)l->data);
+    info->stream_list = g_list_append (info->stream_list, s);
+
+    if (GST_IS_PLAYER_AUDIO_INFO(s))
+      info->audio_cachelist = g_list_append (info->audio_cachelist, s);
```
sdroege: This is not really a "cache", maybe just call it audio_stream_list (same for the other two)

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1165,514 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+
+  g_signal_emit (data->player, signals[SIGNAL_MEDIA_INFO_UPDATED], 0, data->info);
+
+  return FALSE;
+}
+
+static void
+free_media_info_updated_signal_data (MediaInfoUpdatedSignalData *data)
+{
+  GST_DEBUG_OBJECT (data->player, "media_info %p", data->info);
```
sdroege: No debug output needed here IMHO

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1165,514 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    data->info = gst_player_media_info_copy (self->priv->media_info);
+    g_mutex_unlock (&self->priv->lock);
+
+    GST_DEBUG_OBJECT (self, "media_info %p", data->info);
+
+    g_main_context_invoke_full (self->priv->application_context,
+        G_PRIORITY_DEFAULT, media_info_updated_dispatch,
+        data, (GDestroyNotify) free_media_info_updated_signal_data);
+  } else {
+    g_signal_emit (self, signals[SIGNAL_MEDIA_INFO_UPDATED], 0);
```
sdroege: You have to provide the info in g_signal_emit() here too. This should be: lock(); info=copy_info(); unlock(); emit(); unref(info);

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1165,514 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+{
+  if (self->priv->dispatch_to_main_context) {
+    MediaInfoUpdatedSignalData *data = g_slice_new
+                                        (MediaInfoUpdatedSignalData);
+    data->player = self;
+    g_mutex_lock (&self->priv->lock);
+    data->info = gst_player_media_info_copy (self->priv->media_info);
+    g_mutex_unlock (&self->priv->lock);
+
+    GST_DEBUG_OBJECT (self, "media_info %p", data->info);
```
sdroege: Not really useful debug output :)

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1165,514 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  GstPlayerStreamInfo *stream_info)
+{
+  GstPlayerSubtitleInfo *info = (GstPlayerSubtitleInfo*) stream_info;
+
+  if (stream_info->tags) {
+
+    if (info->language) {
+      g_free (info->language);
+      info->language = NULL;
+    }
```
sdroege: Maybe we should also do that if there are no tags anymore? Also the resetting to defaults in general

lib/gst/player/gstplayer.c
```diff
@@ -1121,6 +1165,514 @@ element_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+    s = gst_caps_get_structure (stream_info->caps, 0);
+    if (s) {
+      gint width, height;
+      gint fps_n, fps_d;
+      gint par_n, par_d;
+
+      if (gst_structure_get_int (s, "width", &width))
+        info->width = width;
+      else
+        info->width = -1;
```
sdroege: Like also here, if there are no caps anymore maybe it should also be reset to -1

sdroege: Which basically means that we could always create the info from scratch if it changed

sdroege: The stream info, not the complete media info I mean

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * Returns: (transfer full): Current media info instance.
+ * the caller should free it with g_object_unref()
+ */
+GstPlayerMediaInfo*
+gst_player_get_media_info (GstPlayer *self)
+{
+  GstPlayerMediaInfo *info;
+
+  g_mutex_lock (&self->priv->lock);
+  info = gst_player_media_info_copy (self->priv->media_info);
```
sdroege: priv->media_info() might be NULL

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * gst_player_get_media_info:
+ * @player: #GstPlayer instance
+ *
+ * Returns: (transfer full): Current media info instance.
+ * the caller should free it with g_object_unref()
+ */
+GstPlayerMediaInfo*
+gst_player_get_media_info (GstPlayer *self)
+{
+  GstPlayerMediaInfo *info;
```
sdroege: g_return_val_if_fail(GST_IS_PLAYER(self), NULL);

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * @player: #GstPlayer instance
+ *
+ * Returns: (transfer full): current audio track
+ * the caller should free it with g_object_unref()
+ */
+GstPlayerAudioInfo*
+gst_player_get_current_audio_track (GstPlayer *self)
+{
+  GstPlayerAudioInfo *info;
+
```
sdroege: g_return_val_if_fail();

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+ * gst_player_get_current_video_track:
+ * @player: #GstPlayer instance
+ *
+ * Returns: (transfer full): current video track
+ * the caller should free it with g_object_unref()
+ */
+GstPlayerVideoInfo*
+gst_player_get_current_video_track (GstPlayer *self)
+{
+  GstPlayerVideoInfo *info;
```
sdroege: g_return_val_if_fail();

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+/**
+ * gst_player_get_current_subtitle_track:
+ * @player: #GstPlayer instance
+ *
+ * Returns: (transfer none): current subtitle track
+ */
+GstPlayerSubtitleInfo*
+gst_player_get_current_subtitle_track (GstPlayer *self)
+{
+  GstPlayerSubtitleInfo *info;
```
sdroege: g_return_val_if_fail();

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+/**
+ * gst_player_set_audio_track:
+ * @player: #GstPlayer instance
+ * @stream_index: stream index
+ */
+gboolean
+gst_player_set_audio_track (GstPlayer *self, gint stream_index)
+{
+  GstPlayerStreamInfo *info;
+
```
sdroege: g_return_val_if_fail();

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+/**
+ * gst_player_set_video_track:
+ * @player: #GstPlayer instance
+ * @stream_index: stream index
+ */
+gboolean
+gst_player_set_video_track (GstPlayer *self, gint stream_index)
+{
+  GstPlayerStreamInfo *info;
+
```
sdroege: etc

lib/gst/player/gstplayer.c
```diff
@@ -1660,6 +2229,204 @@ gst_player_get_pipeline (GstPlayer * self)
...
+  g_mutex_lock (&self->priv->lock);
+  info = gst_player_stream_info_find (self, self->priv->media_info,
+            GST_TYPE_PLAYER_AUDIO_INFO, stream_index);
+  g_mutex_unlock (&self->priv->lock);
+  if (!info) {
+    GST_ERROR_OBJECT (self, "invalid audio stream index %d", stream_index);
+    return FALSE;
+  }
+
```
sdroege: You have to unref() the info here (and in the other functions)

sdroege:
Some more minor things, but nothing big anymore :) I think next version can be merged

gst-play/gst-play.c
codomania: How about if I print media information once ?

e.g something like
static void media_info_cb (GstPlayer *player, GstPlayerMediaInfo *inf, GstPlay *play)
{
  static int once = 0;
  if (!once) {
   call function to print the media information
   once=1;
  }
}

sdroege: Sure, that also works

lib/gst/player/gstplayer.c
codomania: just curious, why do I need to unref() info ? The gst_player_stream_info_find() will return the location of the streamInfo in our streams_list. We don't call g_object_ref() inside the stream_find() function to increase the ref count. The function is an internal function. the goal of function is so check if given stream_index exist in our stream_list.

codomania:
great thanks for review. yes now we are getting very close :) Do you also want me to take care of documentation part (i.e adding functions and structures in docs/lib/gst-player-sections.txt so that media info function shows up in html docs) in next version ?

sdroege:
Yes please :)

lib/gst/player/gstplayer.c
sdroege: Hmm good point, ignore this comment. Sorry!

sdroege:
Replaced by https://github.com/sdroege/gst-player/pull/21

PR #21, opened 2015-04-22
mediainfo-v6

codomania:
Changes since v5:

* added media-info api in doc section. 
* update gst-play to print media information once in media_info_cb
* assertion check in gst_player_get_media_info(), gst_player_get_current_<*>_track ()
* create and emit media_info in g_emit_signal ()
* if tags or caps are not available in the stream then reset the values to default.
* rename "*_cachelist" to "*_stream_list"

I have done some basic test using the media files from tests/media directory and also Sintel_DivXPlus_6500kbps.mkv. The only thing which I could not test are:
gst_player_set_<*>_track()
gst_player_<*>_track_enabled()

because I do not have changes in gst-play to handle these yet. I think we should be able to validate these API() as part of gtk-play changes (which will a bit easier to select the input stream etc...)


sdroege:
Thanks, merged :)

Are you planning to create pull requests for the GTK stuff too? And do you have any other changes planned/pending?

codomania:
Thanks !

Yes I will create pull request for GTK stuff, will try to break down in the following patches

* add a information button next to volum slider, clicking this button will open media information window (which will display all the media details)
* if we are playing only audio then display GST_TAG_IMAGE in video area.
* add a popup menu which will have the following
  * audio track list and selection
  * subtitle track list and selection
  * video track list and also make options to change the pixel-aspect-ratio
* add external subtitle support in player
* extend popup menu list to add option to browse and accept new file for subtitle.

I am new to GTK world, so learning as I go :)

sdroege:
Sounds great, thanks :)

sdroege:
Some tests would probably be also nice, for checking if all this stuff continues to work

codomania:
yes agreed. Will add that in my todo list.

PR #22, opened 2015-04-23
Make sure dimensions are valid

jberkel:
Layouting code crashes if 0/0 get passed

just a workaround, looks like there is a problem in gstreamer itself


sdroege:
No, the "problem" is in GstPlayer :) We notify this also for audio streams, so that when you switch from video to audio-only, you know that by checking for width==height==0.

The video-size-changed signal will disappear soonish and be replaced by proper media information handling, where you won't have this problem anymore.

But merged this until then :)

PR #24, opened 2015-04-24
Rishabh7m patch 6

rishabh7m:
gst_player_set_playback_rate() uses gst_player_seek()

### DIFF
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -724,7 +724,7 @@ void gst_player_set_playback_rate (GstPlayer * self, gdouble rate)
 {
   g_return_if_fail (GST_IS_PLAYER (self));
   self->priv->rate = rate;
-  g_main_context_invoke (self->priv->context, gst_player_set_playback_rate_internal, self);
+  gst_player_seek (self, 0);
 }
 
 typedef struct
@@ -1445,7 +1445,7 @@ gst_player_stop (GstPlayer * self)
 static void
 gst_player_seek_internal_locked (GstPlayer * self)
 {
-  GstClockTime position;
+  GstClockTime position,start_pos;
   gboolean ret;
   GstStateChangeReturn state_ret;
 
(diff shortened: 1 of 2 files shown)


sdroege:
Please squash all the commits together into one

lib/gst/player/gstplayer.c
```diff
@@ -1443,10 +1450,18 @@ gst_player_seek_internal_locked (GstPlayer * self)
 
   remove_tick_source (self);
   self->priv->is_eos = FALSE;
+  
+  start_pos = gst_player_get_position(self);
```
sdroege: This now breaks seeks, you would now always seek to the current position. The seek function needs some refactoring to handle both cases.

lib/gst/player/gstplayer.c
```diff
@@ -688,6 +688,13 @@ eos_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
   self->priv->is_eos = TRUE;
 }
 
+void gst_player_set_playback_rate (GstPlayer * self, gdouble rate)
+{
+  g_return_if_fail (GST_IS_PLAYER (self));
+  self->priv->rate = rate;
+  gst_player_seek (self, 0);
```
ened: Is this for playback speed? If so, shouldn't it continue at the same position?

lib/gst/player/gstplayer.c
```diff
@@ -688,6 +688,13 @@ eos_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
   self->priv->is_eos = TRUE;
 }
 
+void gst_player_set_playback_rate (GstPlayer * self, gdouble rate)
```
ened: Should we add this to `gstplayer.h`?

lib/gst/player/gstplayer.c
sdroege: Yes, see my comments. This all has to be done differently and refactored to have unified seeking logic for play speed and seeking.

sdroege:
This was now solved by another pull request

PR #26, opened 2015-04-25
gtk-play: add media information window 

codomania:
Version 1.0:

* add media information window
* update the player title to stream title

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,7 +53,7 @@ AC_SUBST(LT_AGE)
 PKG_PROG_PKG_CONFIG
 
 PKG_CHECK_MODULES(GLIB, [glib-2.0 gobject-2.0])
-PKG_CHECK_MODULES(GSTREAMER, [gstreamer-1.0 >= 1.4 gstreamer-video-1.0 >= 1.4 gstreamer-tag-1.0 >= 1.4])
+PKG_CHECK_MODULES(GSTREAMER, [gstreamer-1.0 >= 1.4 gstreamer-video-1.0 >= 1.4 gstreamer-tag-1.0 >= 1.4 gstreamer-pbutils-1.0 >= 1.4])
 
 GLIB_PREFIX="`$PKG_CONFIG --variable=prefix glib-2.0`"
 AC_SUBST(GLIB_PREFIX)
--- a/docs/lib/gst-player-sections.txt
+++ b/docs/lib/gst-player-sections.txt
@@ -87,6 +87,7 @@ gst_player_media_info_get_stream_list
 gst_player_stream_info_get_index
 gst_player_stream_info_get_caps
 gst_player_stream_info_get_tags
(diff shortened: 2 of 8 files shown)


sdroege:
Looks good, but it crashes for me on every video I tried. Backtrace here: https://gist.github.com/sdroege/a2d75d5859466e0416f0
Note the type and label in stack frame 6. I didn't look closer yet

sdroege:
On audio-only files it seems to work without crashes, but there are many g_criticals() when closing it

codomania:
Will take a look thanks. I tried couple of video and it worked for me. Does
those video work okay with gst-play?

codomania:
There was one case which could cause failure, please try the above patch and let me know if it works for you ? I tried some media from http://samplemedia.linaro.org/H264/ and they seem to work okay (with assertion or fault) .

sdroege:
Works, thanks :) I'll merge it in a bit.

If there is no codec tag, you could use gst_pb_utils_get_codec_description() on the caps. Also there's GST_TAG_CONTAINER_FORMAT, GST_TAG_AUDIO_CODEC, GST_TAG_VIDEO_CODEC, GST_TAG_SUBTITLE_CODEC. Maybe we should have some char* codec (and container format) description in GstPlayerMediaInfo / GstPlayerStreamInfo instead of letting the app do that?

codomania:
Let me add api in player to get title, codec etc info then will use those
in player. Drop this patch for now.

sdroege:
Ok. Also note that you can just update pull requests, you don't have to create new ones.

For that just push new commits to the branch you used to create the pull request. Also feel free to squash together, reorder, drop commits as needed (i.e. push --force the changes after rebase, etc).

codomania:
okay regarding making transfer none. I was thinking to do that but one question came to mind. if media has more than one stream, is it possible to have different title name for each track ?

sdroege:
Yes

codomania:
so if we have more than one title in the stream, which title should we return from gst_player_media_info_get_title () ?  For now my approach was simple, I just return the title from from first available stream.

sdroege:
For tags we should also handle the TAG message, and in there we can get told if the taglist is of global scope or not. These tags are the ones that should go into the GstPlayerMediaInfo.

For the title, you could probably do some heuristic there. If it's not in the GstPlayerMediaInfo tags, you get the one from the first video stream you find that has a title, and if no video from the first audio stream.


Maybe split the title part from the codec name part. It's a completely different story :)

codomania:
yes make sense :) lets handle title in different commit.

codomania:
updated with your review feedback.

codomania:
okay let me update the API name .. and fix typo

codomania:
updated the patch.

sdroege:
Thanks, merged :)

PR #27, opened 2015-04-26
player: add API to get global tag

codomania:


codomania:
I have pushed 4 commits on this branch to add the following new API's:
* gst_player_media_info_get_tags
* gst_player_media_info_get_title
* gst_player_media_info_get_container_format
* gst_player_media_info_get_cover_image

Let me know if I am on right path.

lib/gst/player/gstplayer-media-info-private.h
```diff
@@ -94,7 +94,11 @@ struct _GstPlayerMediaInfo
   GObject parent;
 
   gchar *uri;
+  gchar *title;
+  gchar *container;
   gboolean seekable;
+  GstTagList *tags;
+  GstBuffer *cover_image;
```
sdroege: Store the GstSample here. You don't need just the buffer but also the caps, and the info struct in the sample contains information if it's the front cover, back cover, etc.

lib/gst/player/gstplayer-media-info.c
```diff
@@ -540,6 +552,14 @@ gst_player_media_info_copy (GstPlayerMediaInfo * ref)
   info = gst_player_media_info_new (ref->uri);
   info->duration = ref->duration;
   info->seekable = ref->seekable;
+  if (ref->tags)
+    info->tags = gst_tag_list_copy (ref->tags);
```
sdroege: You can just take a new ref for the tag lists btw, also elsewhere where copying

lib/gst/player/gstplayer-media-info.c
```diff
@@ -540,6 +552,14 @@ gst_player_media_info_copy (GstPlayerMediaInfo * ref)
...
+  if (ref->title)
+    info->title = g_strdup (ref->title);
+  if (ref->container)
+    info->container = g_strdup (ref->container);
+  if (info->cover_image)
+    info->cover_image = gst_buffer_copy (info->cover_image);
```
sdroege: And for the buffer/sample too, and for the caps

lib/gst/player/gstplayer.c
```diff
@@ -1226,6 +1227,36 @@ request_state_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+  GstTagList *tags = NULL;
+
+  gst_message_parse_tag (msg, &tags);
+
+  if (gst_tag_list_get_scope (tags) == GST_TAG_SCOPE_GLOBAL) {
+    g_mutex_lock (&self->lock);
+    if (self->media_info) {
+      if (self->media_info->tags) {
+        gst_tag_list_unref (self->media_info->tags);
+        self->media_info->tags = tags;
```
sdroege: I think the last line should be outside the if, otherwise you only set tags if they were previously set

lib/gst/player/gstplayer.c
```diff
@@ -1226,6 +1227,36 @@ request_state_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+      }
+    }
+    else {
+      if (self->global_tags)
+        gst_tag_list_unref (self->global_tags);
+      self->global_tags = tags;
+    }
+    g_mutex_unlock (&self->lock);
+
+    GST_DEBUG_OBJECT (self, "updating global tags");
```
sdroege: You probably want to emit the media-info-updated signal here as the tags have updated (but only if we had a media-info)

lib/gst/player/gstplayer.c
```diff
@@ -1787,6 +1818,105 @@ subtitle_changed_cb (GObject * object, gpointer user_data)
...
+
+  return title;
+}
+
+static void *
+get_container (GstTagList *tags)
+{
+  gchar *container = NULL;
+
+  gst_tag_list_get_string (tags, GST_TAG_CONTAINER_FORMAT, &container);
```
sdroege: As a fallback, you could also get the caps from the source's srcpad (playbin's "source" property). It might contain caps. Or check the file extension in the URI if any. Maybe just add a TODO comment here for the future

lib/gst/player/gstplayer.c
```diff
@@ -1787,6 +1818,105 @@ subtitle_changed_cb (GObject * object, gpointer user_data)
...
+    caps = gst_sample_get_caps (sample);
+    caps_struct = gst_caps_get_structure (caps, 0);
+    gst_structure_get_enum (caps_struct,
+        "image-type", GST_TYPE_TAG_IMAGE_TYPE, &type);
+
+    if ((type == GST_TAG_IMAGE_TYPE_FRONT_COVER) ||
+        (type == GST_TAG_IMAGE_TYPE_BACK_COVER)) {
+      cover_sample = sample;
+    }
+    gst_caps_unref (caps);
```
sdroege: You don't have to unref the caps, they are owned by the sample

lib/gst/player/gstplayer.c
```diff
@@ -1796,6 +1926,7 @@ gst_player_media_info_create (GstPlayer * self)
   GST_DEBUG_OBJECT (self, "begin");
   media_info = gst_player_media_info_new (self->uri);
   media_info->duration = gst_player_get_duration (self);
+  media_info->tags  = self->global_tags;
```
sdroege: And set self->global_tags to NULL. Otherwise you might free them twice.

Also unref global_tags in stop_internal(), finalize() and emit_error()

codomania:
updated patch with your review feedback. I have still kept the as gst_player_media_info_get_cover_image(), let me know if you recommend changing to *_cover_sample () since now we return GstSample*

codomania:
I have tried some media files but none of them contained global tags, do you have any file which contains global tags ? Just wanted to verify the global_tags code path.

sdroege:
Do you get global-scope tags when playing an Ogg, MP4, Matroska file?

codomania:
i do not see global tags. I always get local tags (i have tried ~/Downloads/Sintel_DivXPlus_6500kbps.mkv, ~/Download/300.mp4 tests/media/audio-video.ogg) and several others. I am using 14.10 
brijesh@brijesh-M11BB:~/gst-player$ gst-launch-1.0 --version
gst-launch-1.0 version 1.4.3
GStreamer 1.4.3
https://launchpad.net/distros/ubuntu/+source/gstreamer1.0

sdroege:
I think that's this bug here: https://bugzilla.gnome.org/show_bug.cgi?id=679768

matroska demuxer is sending global tags, and then downstream is breaking it and only sends stream tags

sdroege:
Actually this was a different bug: http://cgit.freedesktop.org/gstreamer/gst-plugins-good/commit/?id=9119fbd774093e3ae762c8652acd80d54b2c3b45 http://cgit.freedesktop.org/gstreamer/gstreamer/commit/?id=18b058100940bdcaed86fa412e3582a02871f995

sdroege:
At least gst-discoverer-1.0 shows the container-format tag now. gst-player still does not show it as part of the per-stream tag lists and also gets no global tag message.

codomania:
in gst-player we are building media info with "audio", "video" and "subtitle" tags received thru playbin and for global tags we are looking into bus tags message callback. And so for my debug indicated that bus tag message callback is getting local tags (same tags we get via <*>_tags_changed signal). In gst-discoverer they have four stream types "audio", "video", "sub" and "container". In our case we are missing "container" stream type. But I am thinking why do we need container stream type, what player would do with container stream type information ? our version of gst_media_info_get_container_format() is taking the following approach to get the container name (just in case player would to get that information):
* look for GST_TAG_CONTAINER_FORMAT in global tag
* if global tag does not exist:
  * try to look for GST_TAG_CONTAINER_FORMAT in video stream tags
  * if not found, try to look for it in audio stream tag.
By using the above approach so far we have managed to get the container format string. 

I am using the same approach for "media_info_get_title()",  "media_info_get_cover_image()" and "media_info_get_container_format()".  In my above patch gst-play prints the container_format string as expected. Only part i wanted to understand is why we never get the global tags in our callback and sounds like its bug.  What's your recommendation, should I build the gst source from latest tree and try our code to verify the global tag path ?

codomania:
ah i just noticed that you made commit today in those repos. I have typically used the gst builds from distro. Let me give a shot to build from git and see if I can work on gst-player to verify the global tag logic.

sdroege:
The "container" stuff in gst-discoverer (and the topology in general) is confusing and a wart in the API :)

What you do in the code is correct, it just doesn't work (you never get global scope tags) because of GStreamer bugs (in specific the one I linked above).

codomania:
the above commit adds support to display cover art (when available) in gtk-play.

codomania:
found small bug in gtk uri list iterator logic, the commit e0b1333b979ab0fea9cbb1bf7289d340ad7078c6 fixes this.

The problem can be reproduced by using the following options
$ ./gtk/gtk-play ~/Download/sample.mp4 ~/Download/sample.mp4 ~/Download/sample1.mp4 
The above args cause next button to never play sample1.mp4 and also prev button causes g_assertion

gtk/gtk-play.c
```diff
@@ -55,6 +59,9 @@ typedef struct
   GtkWidget *media_info;
   gulong seekbar_value_changed_signal_id;
   gboolean playing;
+  GdkPixbuf *cover_pixbuf;
+  GstSample *cover_sample;
+  GMutex lock;
```
sdroege: Why do we need a mutex here? All signals should be emitted from the main thread

gtk/gtk-play.c
```diff
@@ -662,12 +710,121 @@ eos_cb (GstPlayer * unused, GtkPlay * play)
...
+  buffer = gst_sample_get_buffer (sample);
+  caps = gst_sample_get_caps (sample);
+  caps_struct = gst_caps_get_structure (caps, 0);
+
+  gst_structure_get_enum (caps_struct, "image-type",
+      GST_TYPE_TAG_IMAGE_TYPE, &type);
+
+  /* FIXME: Should we check more type ?? */
+  if ((type != GST_TAG_IMAGE_TYPE_FRONT_COVER) &&
+      (type != GST_TAG_IMAGE_TYPE_UNDEFINED)) {
```
sdroege: GST_TAG_IMAGE_TYPE_NONE maybe

gtk/gtk-play.c
```diff
@@ -662,12 +710,121 @@ eos_cb (GstPlayer * unused, GtkPlay * play)
...
+    return NULL;
+  }
+
+  loader = gdk_pixbuf_loader_new ();
+  if (gdk_pixbuf_loader_write (loader, info.data, info.size, &err) &&
+      gdk_pixbuf_loader_close (loader, &err)) {
+    pixbuf = gdk_pixbuf_loader_get_pixbuf (loader);
+    if (pixbuf)
+      g_object_ref (pixbuf);
+    else {
```
sdroege: Add some { and } to the if case

docs/lib/gst-player-sections.txt
```diff
@@ -82,6 +82,11 @@ GstPlayerSubtitleInfo
 
 gst_player_media_info_get_uri
 gst_player_media_info_get_duration
+gst_player_media_info_get_title
+gst_player_media_info_get_container_format
+gst_player_media_info_is_seekable
+gst_player_media_info_get_cover_image
```
sdroege: Cover image seems to specific for the API here IMHO. Should be something more generic

gtk/gtk-play.c
```diff
@@ -638,15 +678,23 @@ eos_cb (GstPlayer * unused, GtkPlay * play)
...
       gtk_widget_set_sensitive (play->next_button, g_list_next (next) != NULL);
 
       gtk_widget_set_sensitive (play->media_info, FALSE);
+      g_mutex_lock (&play->lock);
+      if (play->cover_pixbuf)
+        g_object_unref (play->cover_pixbuf);
+      play->cover_pixbuf = NULL;
+      play->cover_sample = NULL;
+      g_mutex_unlock (&play->lock);
+      gtk_widget_hide (play->video_area);
```
sdroege: The hiding and showing all the time will probably cause some flickering

lib/gst/player/gstplayer.c
```diff
@@ -1226,6 +1240,60 @@ request_state_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
 }
 
 static void
+media_info_update (GstPlayer *self, GstPlayerMediaInfo *info)
+{
+  if (info->title) {
+    g_free (info->title);
+    info->title = get_from_tags (self, info, get_title);
```
sdroege: You want to do the assignment outside the if (also for the following lines)

lib/gst/player/gstplayer.c
```diff
@@ -1226,6 +1240,60 @@ request_state_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+{
+  GstPlayer *self = GST_PLAYER (user_data);
+  GstTagList *tags = NULL;
+
+  gst_message_parse_tag (msg, &tags);
+
+  GST_DEBUG_OBJECT (self, "recieved %s tags",
+    gst_tag_list_get_scope(tags) == GST_TAG_SCOPE_GLOBAL ? "global":"local");
+
+  if (gst_tag_list_get_scope (tags) == GST_TAG_SCOPE_GLOBAL) {
```
sdroege: Maybe add a comment here linking to the relevant GStreamer bug

lib/gst/player/gstplayer.c
```diff
@@ -1226,6 +1240,60 @@ request_state_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
...
+static void
+tags_cb (GstBus * bus, GstMessage *msg, gpointer user_data)
+{
+  GstPlayer *self = GST_PLAYER (user_data);
+  GstTagList *tags = NULL;
+
+  gst_message_parse_tag (msg, &tags);
+
+  GST_DEBUG_OBJECT (self, "recieved %s tags",
+    gst_tag_list_get_scope(tags) == GST_TAG_SCOPE_GLOBAL ? "global":"local");
```
sdroege: The scopes are GLOBAL and STREAM, not LOCAL :)

gtk/gtk-play.c
codomania: play->cover_pixbuf is accessed by both gtk_widget_queue_draw() and media_info_updated_cb ()  hence I was using mutex. The idea is media_info_updated_cb() updates the pixbuf and draw function draw's the updated picture.

docs/lib/gst-player-sections.txt
codomania: how does gst_player_media_info_get_image_sample () sounds ? this name tells exactly what we are returning :)

gtk/gtk-play.c
sdroege: They're guaranteed both to be called from the main thread only :)

docs/lib/gst-player-sections.txt
sdroege: I don't have a better idea, so let's go with that :)

gtk/gtk-play.c
codomania: ah got it.. should have reminded myself that  its gtk app and not gstplayer lib :)

codomania:
updated the patch with your review feedback.

codomania:
added new patch to add popup menu for track selection. It seems disabling the video is unstable in gstreamer playbin. When i disable the video then audio is also stopped.. it works okay only ximagesink is used but with overlay it causes pipeline to hang. It seems like either playbin has some bug or disabling the video on-fly is not supported feature in gstreamer. let me know your though ?

sdroege:
Thanks, I'll take a look at the new patches after the weekend at latest :)


Supporting the video on-the-fly is supposed to work, and what you observed there is a bug. Currently stream switching and disabling/enabled during playback is not 100% reliable... but having this feature implemented here will give yet another reason to fix that soonish :)

codomania:
okay sure, take your time. Yes now we  have reason to fix on-the-fly disable/enable and switching tracks :). I did some basic test and track selection is working okay, only part is something cover art is not displayed as soon as we draw it.. need to scale or focus the window to redraw it.. looks like i am missing on calling some gtk API to refresh our drawing area.

codomania:
I looked at why my cover art is not getting displayed as soon as I disable the video. It does seem that we are doing right things from application.  In debug, I found that g_object_set (G_OBJECT(playbin), "current-*", <value>) is not a blocking call. So after disabling the video I immediately display the cover art but in background playbin is still removing the video pad and as part of removal of pad imagesink calls its reset function which repaints background in black color and if my cover art is displayed before this painting then it gets repainted.. 

If I delay cover art display by a second then it works as expected. This is hinting us with two choices
- fix playbin to make the flag setting call blocking.
- or run a idle loop in gtk and call cover art display.

sdroege:
Alternatively it should also work to either set the window handle to NULL, or to have two separate widgets (one for the image, one for the video) and hide/show them accordingly. By giving away the window handle you basically give away all control of the window and can't really do anything with it anymore.

codomania:
Yes we can do that also but like you pointed in previous patch that disabling and enabling the widgets will cause a small flicker. It's not that bad looking flicker but still. So in my patch i did implemented a simple idle loop to queue any draw event (if video or subtitle is not active). That will ensure that whatever pixbuf we render will be eventually get display. I am open to any of the options you would like to take..

sdroege:
I think in this case the flickering is ok :) But maybe take a look at how totem manages this

codomania:
:) I looked at totem and unfortunately it does not have support enabling/disabling the tracks. Looks at this hinting me that we have much more features than exported by totem :)  I looked at totem menu and all i see is options to choose different subtitle tracks but other than it does not have option to choose different audio track or video tracks or disable audio or video track.

I will go ahead and implement two different widget (one for video and one for image) and will show and hide the widget accordingly. I have tried it and it does not look bad at all. it's pretty smooth transition.

codomania:
So today i upgrade to Ubuntu 15.04 and it came with later version of gtk (i.e gtk 3.15) and gtk_widget_set_double_buffered () is deprecated in this release. It seems like this API is not portable and it may work for now but may not work with Wayland or Mir backend hence recommendation is not to use this. Should I remove this from our gtk app ?

sdroege:
Yes, that can be removed, thanks!

sdroege:
We also use some other deprecated API btw, feel free to clean that up

sdroege:
I merged some of the commits already, everything else is reviewed

codomania:
okay, i am breaking down commits for easy review.

codomania:
in this pull lets review and merge track selection support and once its done then will create a new pull request to handle covert art thing. i will soon incorporate ur feedback on track selection and push updated patch

sdroege:
Sure, whatever works best for you :)

codomania:
okay patches are now ready. I have taken out cover art stuff. Please let me know your review feedback. I think once this patch is merge then we can take a look into gstreamer to see what is happening with disabling tracks on-the-fly and how we can fix it :)

codomania:
BTW, I remember seeing your comment/question on playbin restore flag but somehow its now lost in our thread. Actually in testing i disabled the audio track and then hit next to play next uri and could not hear anything. so that clearly indicated that track is disable. And yes it was.

So I decided to look into other players and see how they work. I did same in VLC and vlc restore to original state when you change to next uri. Hence I thought gstplayer should behave the similar. But i am fine with not restoring it. I guess user can anyway go to track selection menu and can see that track is disabled hence he/she is not seeing/hearing anything.. for now i have dropped that patch from this series.

sdroege:
I don't have a strong opinion on this, it just seems weird if properties change under the hood without the application knowing :) Maybe let's create a ticket about this so we don't forget

codomania:
sure, will create a ticket and we can revisit it later.

sdroege:
Thanks, merged this all :)

PR #28, opened 2015-05-03
build: remove the use of GST_PKG_CONFIG_PATH

ceyusa:
In gstreamer, that variable is defined in common/ submodule. But it is not
used here. As GST_PKG_CONFIG_PATH is not defined, it resets the
PKG_CONFIG_PATH breaking the compilation in jhbuild.

This patch removes the PKG_CONFIG_PATH assignation when launching the gir
scanner.


sdroege:
Thanks, merged :)

PR #31, opened 2015-05-05
Use gradle to build the Android App 

ened:
Also added a file browser so that users can select movies detected by Android MediaStore.

device-2015-05-05-160727


sdroege:
Thanks, looks good from a short review. This was on my todo list since quite some time already.
I'll take a closer look later and then merge :)

ened:
I got more! Once merged, we can start doing extensions to the player and add a proper UI. :-)

wandering7man:
Nice ;)

sdroege:
Oh yes, a proper UI would be awesome. Same for the iOS and GTK UI, or some to-be-written Qt UI. I'm not an UI person ;)

wandering7man:
Any ideas guys, why video streaming doesn't work on 2k+ screen resolution?:(
1024x768 work great;)

ened:
@sdroege with all the buzz around material design we should stick to it as close as possible, hence the Android player should probably be based on a standard Android Java stack.

sdroege:
@wandering7man maybe because of https://bugzilla.gnome.org/show_bug.cgi?id=731204 , but please report GStreamer related bugs there and not here please :)

@ened Sure

android/app/build.gradle
```diff
@@ -0,0 +1,66 @@
+apply plugin: 'com.android.application'
+
+android {
+    compileSdkVersion 22
+    buildToolsVersion "22.0.1"
+
+    defaultConfig {
+        applicationId "org.freedesktop.gstreamer.play"
+        minSdkVersion 19
+        targetSdkVersion 22
```
sdroege: Why such high versions? It should in theory all work with version 9

android/app/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+    // Thank you http://stackoverflow.com/q/28878689/375209
+    tasks.withType(JavaCompile) {
+        compileTask -> compileTask.dependsOn ndkBuild
+    }
+
+    task ndkBuild(type: Exec) {
+        Properties properties = new Properties()
+        properties.load(project.rootProject.file('local.properties').newDataInputStream())
+        def ndkDir = properties.getProperty('ndk.dir')
+        environment GSTREAMER_SDK_ROOT_ANDROID: properties.getProperty('gstreamerSdk.dir')
```
sdroege: The variable names are supposed to be GSTREAMER_ROOT_ANDROID, without the SDK part :) And I guess as such it should also be gstreamer.dir and not gstreamerSdk.dir

android/app/src/main/java/org/freedesktop/gstreamer/play/VideoSelector.java
```diff
@@ -0,0 +1,142 @@
+package org.freedesktop.gstreamer.play;
+
```
sdroege: You might want to add some copyright/license header at the top of the file

android/app/src/main/jni/Android.mk
```diff
@@ -2,17 +2,23 @@ LOCAL_PATH := $(call my-dir)
...
 LOCAL_SHARED_LIBRARIES := gstreamer_android
 LOCAL_LDLIBS := -llog -landroid
+LOCAL_CFLAGS := -I$(GST_PATH)/lib/
 include $(BUILD_SHARED_LIBRARY)
 
 ifndef GSTREAMER_ROOT
-ifndef GSTREAMER_ROOT_ANDROID
-$(error GSTREAMER_ROOT_ANDROID is not defined!)
+ifndef GSTREAMER_SDK_ROOT_ANDROID
+$(error GSTREAMER_SDK_ROOT_ANDROID is not defined!)
```
sdroege: Same here, GSTREAMER_ROOT_ANDROID and not the SDK things.

sdroege:
Looks good except for some cosmetic things. Can you update the pull request accordingly, then I'll merge it :)

android/app/build.gradle
ened: You are right, it is too high.
I suggest to use 16 for the App, mostly because I do not have any older devices. ;-)
Anything lower will catch only about  < 10% of the devices in the market (according to https://developer.android.com/about/dashboards/index.html).

ened:
Done.

1. Do you have any artwork for the App icon? Anything we can remotely use? ;-)
2. I'm not sure if we need to keep maintaining `android/Makefile.am`, any thoughts?

ened:
& please let me know if you want all those commits squashed.

sdroege:
I'll squash them all together when merging if that makes sense, don't worry about that :) But your changes can't be merged with latest master it seems.

I don't have any artwork, no, but you could use the GStreamer logo probably: http://gstreamer.freedesktop.org/artwork/

For android/Makefile.am, it would be useful to keep that updated. Then one could just run "make dist" and get all the sources in a tar.gz, including the Android sources.

ened:
OK done. Let's put the app icon in the parking lot, it requires some dark pixel magic to get the logo ready for Android.

ened:
ic_launcher

Maybe this one. :)

android/app/src/main/jni/Application.mk
```diff
@@ -0,0 +1,2 @@
+APP_ABI := armeabi
+APP_PLATFORM := android-19
```
sdroege: Here it still says android-19

android/app/src/main/jni/Application.mk
```diff
@@ -0,0 +1,2 @@
+APP_ABI := armeabi
```
sdroege: Also add armeabi-v7a and x86 here :)

And inside the Android.mk you would then use something like this, and set those three variables:

ifeq ($(TARGET_ARCH_ABI),armeabi)
GSTREAMER_ROOT        := $(GSTREAMER_ROOT_ARM)
else ifeq ($(TARGET_ARCH_ABI),armeabi-v7a)
GSTREAMER_ROOT        := $(GSTREAMER_ROOT_ARMV7)
else ifeq ($(TARGET_ARCH_ABI),x86)
GSTREAMER_ROOT        := $(GSTREAMER_ROOT_X86)
else
$(error Target arch ABI not supported)
endif

android/app/build.gradle
```diff
@@ -0,0 +1,66 @@
+apply plugin: 'com.android.application'
+
+android {
+    compileSdkVersion 22
+    buildToolsVersion "22.0.1"
+
+    defaultConfig {
+        applicationId "org.freedesktop.gstreamer.play"
+        minSdkVersion 10
```
sdroege: Why not 9? :)

android/app/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+        targetSdkVersion 22
```
sdroege: What is the difference between minSdkVersion and targetSdkVersion? Will the result run on android-9/10 or will it require setting the targetSdkVersion to those first?

ened: From the Android docs about **targetSdk**:
This attribute informs the system that you have tested against the target version and the system should not enable any compatibility behaviors to maintain your app's forward-compatibility with the target version. The application is still able to run on older versions (down to minSdkVersion).

sdroege: Makes sense

android/app/src/main/jni/Application.mk
ened: Should be `GSTREAMER_ROOT_ARM` -> `GSTREAMER_ROOT_ANDROID_ARM`. So we always build all architectures, good approach I think. :)

sdroege:
Thanks, merged :)

PR #32, opened 2015-05-05
Added methods to retrieve current video/audio/subtitle track status

ened:
Added methods:
    gst_player_get_video_track_enabled
    gst_player_get_audio_track_enabled
    gst_player_get_subtitle_track_enabled
    (internal) player_get_flag

These methods can be checked whether a those tracks are actually used.


codomania:
Looks like you are interested in knowing if current audio/video/subtitle track is enabled, if so then app can get the similar info using the below code without adding new API. 
```c
is_audio_track_enabled (GstPlayer *player)
{
     GstPlayerAudioInfo *info = gst_player_get_current_audio_track (player);
     if (!info)
         return FALSE;  /* track is not enabled */
     g_object_unref (info);
     return TRUE;  /* track is enabled */
}
```

ened:
Hi @codomania very interesting and would probably save a few lines. Thanks.

How does it work together with the other method (player_set_flag / player_clear_flag) though?
Where is the real status stored? Once the flag is cleared, will the `gst_player_get_current_audio_track` return null, for example?

codomania:
The flag is not stored in gstplayer libs, based on user input we configure the flags in playbin. 
e.g if user wants to enable/disable the tracks then app will call gst_player_set_*_track_enable (player, TRUE/FALSE). And based on this player_set_flag or player_get_flag will set the corresponding bits in playbin's flag property.  The call to gst_player_get_current_<*>_track() will first check if track is enable and if not then return NULL else return you the track stream info object.

ened:
@codomania hm, then I'd actually say `gst_player_get_video_track_enabled` should work like proposed in the PR. Using `gst_player_get_current<>_track()` to provide a flag whether the track is enabled is then ambiguous, isn't it?

At this level of API, developers might want to know:
1. Is the track enabled?
2. If the track is enabled, is it available?

codomania:
To my understanding, a media can contain multiple tracks of the same type but it will have only one track enabled at a given point of time and that track is referred as current (or active) track. Enable and disable is done on current track only.
 
So when you say: <i>Is the track enabled ?</i> I think you are essentially asking,  does media have an<b>current (or active) </b> track? If so, then gst_player_get_current_track() API is giving  exactly same information. If the track is not active (i.e disabled) then it returns NULL else return StreamInfo object.

sdroege:
Yes, I agree with @codomania here

ened:
Hi both & ok. I'll close this PR then and continue on another part. :)

PR #35, opened 2015-05-07
gtk-play: display cover art

codomania:
We maintain two drawing widgets, image and video. Cover art is drawn in
image widget and video is rendered in video widget. Based on the following
conditions we show either image or video widget:

- if media info does not have active video stream then hide video widget
  and show image widget.
- if media info contains active video stream then show video widget and
  hide the image widget.


sdroege:
Thanks! Didn't you also have a patch to add a fullscreen button and repeat?

sdroege:
Merged

codomania:
Yes I did have both of these patch. Need to rebase with master and push. in my limited testing i did found issue with repeat. It was not gtk issue, instead it seem like our player lib issue. if we repeat the list then sometime pipeline gets stuck during repeating. will take a look and give u update later.

PR #40, opened 2015-05-08
Add repeat and fullscreen support in gtk-play

codomania:
Ignore my comment on other pull request regarding looping have some issue. it turned out to be a issue with our gtk-play seekbar range setting. Now loop and fullscreen is working as expected


sdroege:
Thanks, merged! :)

Wishlist: in fullscreen mode, the controls should disappear after a few seconds and appear again when moving the mouse

PR #42, opened 2015-05-08
gtk-play: autohide control bar in fullscreen mode.

codomania:
fulfilling wishlist :) 

A simple toolbar autohide, animation can be implemented later :) I was thinking to start external subtitle support next

### DIFF
--- a/gtk/gtk-play.c
+++ b/gtk/gtk-play.c
@@ -59,11 +59,14 @@ typedef struct
   GtkWidget *media_info_button;
   GtkWidget *repeat_button;
   GtkWidget *fullscreen_button;
+  GtkWidget *toolbar;
+  GdkCursor *default_cursor;
   gulong seekbar_value_changed_signal_id;
   GdkPixbuf *image_pixbuf;
   gboolean playing;
   gboolean loop;
   gboolean fullscreen;
+  gint toolbar_hide_timeout;
 } GtkPlay;
 
 enum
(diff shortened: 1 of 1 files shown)


gtk/gtk-play.c
```diff
@@ -1253,6 +1299,8 @@ main (gint argc, gchar ** argv)
   gst_player_play (play.player);
   play.current_uri = g_list_first (play.uris);
 
+  g_timeout_add_seconds (5, (GSourceFunc) gtk_hide_toolbar_loop, &play);
```
sdroege: You should add/remove the timeout based on mouse activity. Otherwise the toolbar will be hidden 0-5 seconds after the mouse movement stopped, maybe really immediately.

sdroege: See g_source_remove() and the return value of g_timeout_add_seconds() for removing btw

codomania: Here is why i have gtk_hide_toolbar_loop() running all the time, I was trying to handle the case where u go to full screen and then don't do any mouse activity, in that case we should simply hide the control. And similarly if you move the mouse and then stop moving  then also toolbar should get auto hide (maybe after 5sec ?). I think one thing which I should have done is track of keep track of time for last mouse movement and in gtk_hide_toolbar_loop() check if the time elapsed between last movement and the callback is greater than 5sec then hide the toolbar..
 Or i might be making is unnecessary complex ;)

codomania:
updated to use g_source_remove ()..

codomania:
now also hide mouse pointer in fullscreen mode.

sdroege:
Thanks, merged :)

PR #45, opened 2015-05-11
add external subtitle

codomania:
The patch adds external subtitle support. First patch are some cleanup code in gtk-play. There are two things i noticed (not sure if its bug or playbin limitations)

- playbin does not support adding external subtitle on-fly, hence we need to restart the pipeline after setting the uri. gtk-play takes care of this. 
- playbin (i.e subparse) not able to parse the language of external subtitle hence gst_player_subtitle_info_get_language() returns "NULL".
- it will be nice to know if there is any way (or tag flags) which we can use to identify the external vs in-stream subtitle.

### DIFF
--- a/docs/lib/gst-player-sections.txt
+++ b/docs/lib/gst-player-sections.txt
@@ -52,6 +52,10 @@ gst_player_get_current_subtitle_track
 gst_player_set_audio_track_enabled
 gst_player_set_video_track_enabled
 gst_player_set_subtitle_track_enabled
+
+gst_player_set_subtitle_uri
+gst_player_get_subtitle_uri
+
 <SUBSECTION Standard>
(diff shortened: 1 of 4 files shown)


codomania:
sorry hacked the existing PR, renamed the header and description.. was lazy to create new branch and send new PR ;)

codomania:
updated the patch set to reflect your review feedback.

sdroege:
Can you update the code for the subtitles like I mentioned above? Even if that doesn't work, I'll take a look then and try to make it work. That's how it's supposed to work :)

sdroege:
Oh nevermind, this was done in https://github.com/sdroege/gst-player/pull/48 already which contains this pull request and other changes. Let's close this one then

PR #48, opened 2015-05-14
add visualization support

codomania:
please review the visualization patches and let me know if I am on right track. I could not find any easy way to query the list of visualization element in gstreamer hence used a bit unconventional method (i.e read the plugin metadata kclass and search for word "visualization") to get list of visualization element.

The code works like this:
- during init player create new audio-visualization object.
- application can call gst_player_get_visualization_list() to get list of available gstreamer visualization element.
- application can call gst_player_set_visualization (player, name) to set the visualization element and similarly gst_player_set_visualization_enabled(player, boolean) can be used for enabling and disabling the visualization.

The code is working and I am able to get the visualization and able to change to various elements etc but like video, audio and subtitle disabling and enabling on-the-fly does not work.

### DIFF
--- a/docs/lib/gst-player-sections.txt
+++ b/docs/lib/gst-player-sections.txt
@@ -56,6 +56,12 @@ gst_player_set_subtitle_track_enabled
 gst_player_set_subtitle_uri
 gst_player_get_subtitle_uri
 
+gst_player_get_visualization_elements_name
+gst_player_get_visualization_elements_description
+gst_player_get_current_visualization
+gst_player_set_visualization
+gst_player_set_visualization_enabled
+
 <SUBSECTION Standard>
 GST_IS_PLAYER
 GST_IS_PLAYER_CLASS
--- a/docs/lib/gst-player-sections.txt
+++ b/docs/lib/gst-player-sections.txt
@@ -52,6 +52,10 @@ gst_player_get_current_subtitle_track
 gst_player_set_audio_track_enabled
 gst_player_set_video_track_enabled
 gst_player_set_subtitle_track_enabled
+
+gst_player_set_subtitle_uri
+gst_player_get_subtitle_uri
+
 <SUBSECTION Standard>
 GST_IS_PLAYER
 GST_IS_PLAYER_CLASS
--- a/gtk/gtk-play.c
+++ b/gtk/gtk-play.c
@@ -753,6 +753,85 @@ track_changed_cb (GtkWidget * widget, GtkPlay * play)
     change_track (play, index, type);
 }
 
+static void
+visualization_changed_cb (GtkWidget * widget, GtkPlay * play)
+{
+  gchar *name;
+
+  if (gtk_check_menu_item_get_active (GTK_CHECK_MENU_ITEM (widget))) {
+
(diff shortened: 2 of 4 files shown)


sdroege:
Thanks, I'll review the other parts tomorrow. Just some comments on the visualization things for now :)

codomania:
thanks, I have updated the patch set.. (now removed gstplayer-visual files).

codomania:
the patches are updated with your review feedback.

codomania:
in some testing i found that if visualization element is set then we do not get end-of-stream message or callback and because of that stop playing the next file from playlist. I enabled the debug log and it seems like gstplayer never gets the EOS hence it does not propagate to user app. Seems like a bug in playbin somewhere.. it happens only if visualization element is set and vis flag is enabled. Should I report this in gstreamer forum? I was just wondering if i need to create a test case to reproduce the issue outside gst-player and then submit that.. thoughts ? Or maybe you already know it and have some patch in queue ;)

sdroege:
Oh interesting, that's definitely a bug in GStreamer and it would be great if you could report that at https://bugzilla.gnome.org . You can easily reproduce that with gst-launch-1.0 playbin flags=audio+vis uri=file:///path/to/some/short/audio/file

codomania:
done https://bugzilla.gnome.org/show_bug.cgi?id=749676

codomania:
fyi, did git bisect and found that the below commit in gst-plugins-base broke the audio+vis support. I have also updated the bug report with this update.  

commit 9f81931716b7f8fd29862f0bfa1a3891cb414bdc
Author: Song Bing <b06498@freescale.com>
Date:   Mon Jan 12 16:08:33 2015 +0800

    streamsynchronizer: Send GAP event to finish preroll when change state from PLAYING to PAUSED

    Wait in the event function when EOS is received until all pads are EOS
    and then forward the EOS event from each pads own event function.

    Also send a new GAP event for EOS pads from the event function whenever
    going from PLAYING->PAUSED by shortly waking up the GCond. This is needed
    to allow sinks to pre-roll again, as they did not receive EOS yet because
    we blocked that, but also will never get data again.

sdroege:
Ok, so the visualization problem should be fixed now. Is there anything else on your side that would prevent these changes from being merged? I'll try to review tonight

codomania:
The patches are good for merge. I did some test and it seems to be working fine. I will also add some unit test but first need to create a small video file which contains multiple tracks etc so that we can use that for testing mediainfo and other features.

codomania:
i tried looping the audio-short.ogg for long time and getting the following message after 10 min of loop

(gtk-play:25751): GStreamer-CRITICAL **: gst_poll_write_control: assertion 'set != NULL' failed

The playback is still continue but the screen is flooded with messages. Not able to see the similar issue with vis disabled. Have updated the bug. Not sure how to reproduce it outside the gst-player.

codomania:
I am able to reproduce the above assertion with small modification in gst-play-1.0 and have reopened the bug with steps to reproduce.

sdroege:
Thanks, merged this with some changes in a few places, also to the API.

PR #49, opened 2015-05-25
player: share textures with other OpenGL contexts

amorenoc:
added an API to set a share context, to e.g. draw video onto an existing GL scene.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,7 +53,7 @@ AC_SUBST(LT_AGE)
 PKG_PROG_PKG_CONFIG
 
 PKG_CHECK_MODULES(GLIB, [glib-2.0 gobject-2.0])
-PKG_CHECK_MODULES(GSTREAMER, [gstreamer-1.0 >= 1.4 gstreamer-video-1.0 >= 1.4 gstreamer-tag-1.0 >= 1.4 gstreamer-pbutils-1.0 >= 1.4])
+PKG_CHECK_MODULES(GSTREAMER, [gstreamer-1.0 >= 1.4 gstreamer-video-1.0 >= 1.4 gstreamer-tag-1.0 >= 1.4 gstreamer-pbutils-1.0 >= 1.4 gstreamer-gl-1.0 >= 1.4])
 
 GLIB_PREFIX="`$PKG_CONFIG --variable=prefix glib-2.0`"
 AC_SUBST(GLIB_PREFIX)
--- a/lib/gst/player/Makefile.am
+++ b/lib/gst/player/Makefile.am
@@ -58,6 +58,7 @@ GstPlayer-@GST_PLAYER_API_VERSION@.gir: $(INTROSPECTION_SCANNER) libgstplayer-@G
 		--pkg gstreamer-video-1.0 \
 		--pkg gstreamer-tag-1.0 \
 		--pkg gstreamer-pbutils-1.0 \
+		--pkg gstreamer-gl-1.0 \
 		--pkg-export gstreamer-player-@GST_PLAYER_API_VERSION@ \
 		--add-init-section="gst_init(NULL,NULL);" \
 		--output $@ \
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -134,8 +134,8 @@ struct _GstPlayer
   GstTagList *global_tags;
   GstPlayerMediaInfo *media_info;
 
-  GstGLContext * other_context;
-  GstElement * gl_elem;
+  GstGLContext *other_context;
+  GstElement *gl_elem;
 
   /* Protected by lock */
   gboolean seek_pending;        /* Only set from main context */
@@ -418,7 +418,7 @@ gst_player_set_property (GObject * object, guint prop_id,
       break;
     case PROP_OTHER_GL_CONTEXT:{
       if (self->other_context)
-        g_object_unref(self->other_context);
+        g_object_unref (self->other_context);
 
       self->other_context = g_value_dup_object (value);
       share_context (self, self->other_context);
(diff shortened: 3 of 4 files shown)


sdroege:
This assumes that the sink will always be glimagesink, no?

I think for these custom video rendering things (GL textures, OSX/iOS CALayer, etc) I would rather create subclasses of GstPlayer that ensure everything is set up correctly.

amorenoc:
I see your point, it was a bad idea to "hardcode" a sink. So other than subclassing GstPlayer, how about exposing playbin's video-sink property?
(I am working on a Qt Quick player using gst-player as backend, so far I just get the pipeline and set the video sink there)
Thanks.

sdroege:
I think specific subclasses for this allow the nicest API for an application here as it requires no special GStreamer knowledge to do that. The subclass would then have an API similar to what you proposed here now.

Alternatively you can already do all this without any changes to GstPlayer at all. You could get the playbin instance from GstPlayer (there's API for that), and then just set the "video-sink" property yourself. But in the long run I'd like to have GstPlayer API for the common use cases, like yours.


Let me know if you have your Qt Quick player in a working state, and if you can share the code or want it to be linked as example usage of GstPlayer, or something like that :)

amorenoc:
I will send you a patch soon, thanks.

sdroege:
Any progress here?

amorenoc:
@sdroege the qt binding I am working on is here: bf75665a09cf6fa40983b739cbbb979577c4bc5c

amorenoc:
screen shot 2015-06-12 at 01 38 00
And this is how it looks so far :)

sdroege:
Oh that's great and looks great :) I hope we can get the GTK (and iOS and Android) one into that state too at some point, but I'm not really good with GUI development

amorenoc:
thanks @sdroege ! Would you like me to send you a pull request? I plan to keep working on it. The player is wrapped in QObject-based class, and the video rendering happens in a another object (QQuickItem), so should be easy to have a Widgets-based UI too.

sdroege:
Sure, please send a pull request when you think it's ready :)

sdroege:
You might also want to make sure it works with GStreamer 1.5.1, there were many changes in the GL support that should make your life easier and make things more flexible. However those changes broke backwards compatibility :)

amorenoc:
I will upgrade to 1.5.1 then thanks :)

amorenoc:
@sdroege I followed suit and tried to compile with the latest released binary for OS X, gstreamer-1.0-devel-1.5.1-x86_64.pkg, but after installing the framework I get a 'file not found #include <gst/gl/gstglconfig.h>'. Am I missing something?

sdroege:
Oh no :( This is fixed now with http://cgit.freedesktop.org/gstreamer/cerbero/commit/?id=90cff7d9d76e78d3e7df7d7d84e7927a50777384 but will only be in the pkg if you build it yourself with latest cerbero now... or the next release. Sorry! If you ask on GStreamer IRC on Freenode, someone can probably give you that file for OSX. I don't have access to my mac currently.

amorenoc:
I will *try* to build GStreamer with cerbero master, thanks :)

sdroege:
Let me know if you need some help, just write a mail

sdroege:
You might want to take a look at the QML video sink that is in gst-plugins-bad now. It does all the GL integration bits for you, so should make your life much much easier :)

amorenoc:
that's great! look into it soon, thanks :)

sdroege:
There is also a GstPlayerVideoRenderer interface now, check the implementation for GTK in the GTK example application. I think with the two interfaces (this one and the signal dispatcher) and the QML video sink, you should be able to make clean Qt bindings.


Having a GstPlayerVideoRenderer for generic GL stuff would be useful to have too though, in addition to a QML one.

amorenoc:
@sdroege I am building GStreamer for OS X using cerbero branch 1.6, and some recipes fail to build, since there is a missing libtool archive: **/Library/Frameworks/GStreamer.framework/Versions/1.0/lib/libiconv.la**
These are the build steps:
* ./cerbero-uninstalled bootstrap
* ./cerbero-uninstalled -c config/osx-x86-64.cbc package -f gstreamer-1.0
Any idea what am I doing wrong?

sdroege:
Since 1.6, we use the libiconv provided by OSX/iOS on the system. You'll have to add it to the xcode project file.

amorenoc:
Thanks. Sorry I am not familiar with cerbero, I can't see any xcode project file.

sdroege:
Oh it's in cerbero! Try the wipe command first, and the package again

amorenoc:
that didn't work, libtool keeps failing on packages required libiconv,  first recipe to fail is freetype.
There is this folder cerbero/ide/xcode/, but that's it. I have search the mailing list for similar issues, but  haven't found anything.

sdroege:
Can you file a bug here: https://bugzilla.gnome.org/enter_bug.cgi?product=GStreamer ? Against cerbero and with the full build log. In theory it should work fine after wiping everything first.

amorenoc:
I could build it without issues starting from scratch, thanks for your help again :)

amorenoc:
@sdroege still can't get it working with qt enabled, but I think I am close :)
I added the configure option `--enable-qt` into recipes/gst-plugins-bad-1.0.recipe and also in the same file:
> self.append_env['QT_CFLAGS'] = ' -I/Users/alex/Qt/5.4/clang_64/lib '
> self.append_env['QT_LIBS'] = " -L/Users/alex/Qt/5.4/clang_64/lib "
But when I check the logs I see this:

> configure: *** checking feature: Qt elements ***
> configure: *** for plug-ins: qt ***
> checking for QT... yes
> Package Qt5Core was not found in the pkg-config search path.
> Perhaps you should add the directory containing `Qt5Core.pc'
> to the PKG_CONFIG_PATH environment variable
> No package 'Qt5Core' found
> Package Qt5Core was not found in the pkg-config search path.
> Perhaps you should add the directory containing `Qt5Core.pc'
> to the PKG_CONFIG_PATH environment variable
> No package 'Qt5Core' found
> checking for moc-qt5... no
> checking for moc... no
> checking for rcc-qt5... no
> checking for rcc... no
> checking for uic-qt5... no
> checking for uic... no
> configure: WARNING: Could not find any Qt Windowing integration
> configure: *** These plugins will not be built: qt

I tried this `os.environ['PKG_CONFIG_PATH'] += ":/Users/alex/Qt/5.4/clang_64/lib/pkgconfig"` but didn't work. So my question now is how can I modify PKG_CONFIG_PATH?

ystreet:
For your issue, adding self.new_env['PKG_CONFIG_PATH'] = os.environ.get('PKG_CONFIG_PATH', None) + ":/Users/alex/Qt/5.4/clang_64/lib/pkgconfig" to the recipe should be enough without setting QT_CFLAGS and QT_LIBS.

FWIW, I compiled the sink manually and don't use/install the framework as follows.

1. cerbero build gst-plugins-bad-1.0
2. cerbero shell
3. cd /cerbero/prefix/sources/darwin_x86_64/gst-plugins-bad-1.0-1.5/
4. apply the patch from https://bugzilla.gnome.org/show_bug.cgi?id=755100
5. head config.log to get the configure command line.
6. ./autogen.sh --noconfigure && PKG_CONFIG_PATH=/Users/alex/Qt/5.4/clang_64/lib/pkgconfig:$PKG_CONFIG_PATH ./configure <configure options from 4>
7. make -C ext/qt
8. cp ext/qt/.libs/libgstqtsink.so /cerbero/prefix/dist/darwin_x86_64/lib/gstreamer-1.0/
9. gst-inspect-1.0 qmlglsink should find the sink and output the usual information.

amorenoc:
thanks @ystreet compiling the sink manually did the job. I needed to apply the patch manually though.
But still haven't got it working, I get an error on both your `example/qt/qml` in the source and my code:
`** (<unknown>:88512): CRITICAL **: gst_gl_context_activate: assertion 'GST_GL_IS_CONTEXT (context)' failed`

amorenoc:
Ok, got it working, I was missing macro definitions here:

```
#if GST_GL_HAVE_WINDOW_COCOA && GST_GL_HAVE_PLATFORM_COCOA && defined (HAVE_QT_MAC)
	if (this->priv->display) {
    platform = GST_GL_PLATFORM_CGL;
    gl_api = gst_gl_context_get_current_gl_api (platform, NULL, NULL);
    gl_handle = gst_gl_context_get_current_gl_context (platform);
    if (gl_handle)
      this->priv->other_context =
          gst_gl_context_new_wrapped (this->priv->display, gl_handle,
          platform, gl_api);
  }
#endif
```

sdroege:
Let's close this then, new code here https://github.com/sdroege/gst-player/pull/110

PR #50, opened 2015-06-02
compiler warnings and documentation

ceyusa:
* gtk/gtk-play.c: remove unused variables
* docs/lib/Makefile.am: remoe init-func since gst_init() is not
  required
* lib/gst/player/Makefile.am: remove init-section since gst_init() is
  not required

silence compiler

### DIFF
--- a/pkgconfig/gstreamer-player-uninstalled.pc.in
+++ b/pkgconfig/gstreamer-player-uninstalled.pc.in
@@ -2,6 +2,8 @@
 prefix=
 exec_prefix=
 libdir=${pcfiledir}/../lib/gst/player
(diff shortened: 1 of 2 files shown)


sdroege:
Can you split this into two commits? One for the init() stuff, one for the other?

Also why is the gst_init() not needed? Something sure have to call it?

ceyusa:
> Can you split this into two commits? One for the init() stuff, one for the other?

Sure.

> Also why is the gst_init() not needed? Something sure have to call it?

As far as I understand (I might be wrong) gst_init() is not required to introspect the code, neither for documentation, nor g-i-r. Also gst_player_new() already calls gst_init()

PR #52, opened 2015-06-03
gtk-play : remove duplicate code

sedman:
gtk-play : remove duplicate code
The code is already called in play_current_uri


sedman:
Thanks for your review.
Yes I agree it will be better to move. 
Give me a moment to fix.

sdroege:
Thanks, merged! :)

PR #53, opened 2015-06-03
player: add volume and mute notifier

codomania:
Please let me know what you think about this patch for adding volume and mute notifier. If I am on right path then i can work on adding the code in gtk-play to verify volume and mute callbacks.

### DIFF
--- a/gtk/gtk-play.c
+++ b/gtk/gtk-play.c
@@ -20,6 +20,7 @@
  */
 
 #include <string.h>
+#include <math.h>
 
 #include <gst/gst.h>
 #include <gst/tag/tag.h>
@@ -1429,6 +1430,25 @@ media_info_updated_cb (GstPlayer * player, GstPlayerMediaInfo * media_info,
   }
 }
 
+static void
+player_volume_changed_cb (GstPlayer * player, GtkPlay * play)
+{
+  gdouble new_val, cur_val;
+
+  cur_val = gtk_scale_button_get_value
+    (GTK_SCALE_BUTTON (play->volume_button));
+  new_val = gst_player_get_volume (play->player);
+
+  if (fabs (cur_val - new_val) > 0.001) {
+    g_signal_handlers_block_by_func (play->volume_button,
+        volume_changed_cb, play);
+    gtk_scale_button_set_value (GTK_SCALE_BUTTON (play->volume_button),
+        new_val);
+    g_signal_handlers_unblock_by_func (play->volume_button,
+        volume_changed_cb, play);
+  }
+}
+
 int
 main (gint argc, gchar ** argv)
 {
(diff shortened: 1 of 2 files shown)


gtk/gtk-play.c
```diff
@@ -1430,6 +1431,33 @@ media_info_updated_cb (GstPlayer * player, GstPlayerMediaInfo * media_info,
...
+        volume_changed_cb, play);
+  }
+
+  return FALSE;
+}
+
+static void
+player_volume_changed_cb (GstPlayer * player, GtkPlay * play)
+{
+  g_idle_add ((GSourceFunc) volume_notify_idle_cb, play);
```
sdroege: We should dispatch to the main context inside GstPlayer for this, not let the application do it. Just like it's done for all the other signals :)

lib/gst/player/gstplayer.c
```diff
@@ -328,6 +330,16 @@ gst_player_class_init (GstPlayerClass * klass)
       g_signal_new ("media-info-updated", G_TYPE_FROM_CLASS (klass),
       G_SIGNAL_RUN_LAST | G_SIGNAL_NO_RECURSE | G_SIGNAL_NO_HOOKS, 0, NULL,
       NULL, NULL, G_TYPE_NONE, 1, GST_TYPE_PLAYER_MEDIA_INFO);
+
+  signals[SIGNAL_VOLUME_CHANGED] =
+      g_signal_new ("volume-changed", G_TYPE_FROM_CLASS (klass),
+      G_SIGNAL_RUN_LAST | G_SIGNAL_NO_RECURSE | G_SIGNAL_NO_HOOKS, 0, NULL,
+      NULL, NULL, G_TYPE_NONE, 0, G_TYPE_INT, G_TYPE_INT);
```
sdroege: You have two int arguments here, but don't use them anywhere else for the signal, also not during signal emission

lib/gst/player/gstplayer.c
```diff
@@ -2118,6 +2170,18 @@ subtitle_tags_changed_cb (GstElement * playbin, gint stream_index,
       GST_TYPE_PLAYER_SUBTITLE_INFO);
 }
 
+static void
+volume_notify_cb (GObject *obj, GParamSpec *pspec, GstPlayer *self)
+{
+  emit_volume_changed (self);
```
sdroege: Just inline those functions here, you're only ever going to call them from one place anyway :)

sdroege:
Looks generally good, just some minor comments. Thanks for the patch :)

gtk/gtk-play.c
codomania: Sure, I can takeout the idle loop and do the slider update inside gstplayer context.

lib/gst/player/gstplayer.c
codomania: initially i though that our signal with contain old and new volume values but then decided that userapp can query those directly via gstplayer apis and simplify the logic :)

gtk/gtk-play.c
sdroege: The loop is the same anyway as the GTK app uses dispatch-to-main-context=TRUE on GstPlayer :)

lib/gst/player/gstplayer.c
sdroege: Makes sense, so just remove those here and then it seems good :)

sdroege:
Thanks, merged :)

PR #55, opened 2015-06-07
pc files: add girdir and typelibdir

MathieuDuponchelle:
That way one can expose a GstPlayer as part of another introspected library more easily :)


sdroege:
Thanks!

PR #57, opened 2015-06-10
add color balance api and dialog

ceyusa:
color balance API in gstplayer and its usage in gtk-play

### DIFF
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -45,6 +45,7 @@
 
 #include <gst/gst.h>
 #include <gst/video/video.h>
+#include <gst/video/colorbalance.h>
 #include <gst/tag/tag.h>
 #include <gst/pbutils/descriptions.h>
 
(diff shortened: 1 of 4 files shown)


lib/gst/player/gstplayer.c
```diff
@@ -3179,6 +3180,84 @@ gst_player_set_visualization_enabled (GstPlayer * self, gboolean enabled)
...
+/**
+ * gst_player_get_color_balance_list_channels:
+ * @player: #GstPlayer instance
+ *
+ * Retrieve a list of the available channels.
+ *
+ * Returns: (element-type Gst.ColorBalanceChannel) (transfer none): A
+ *          GList containing pointers to #GstColorBalanceChannel
+ *          objects. The list is owned by the #GstColorBalance
+ *          instance and must not be freed.
```
sdroege: This would mean that the list is owned by playbin, and as such is valid as long as the GstPlayer instance is? Or would it suddenly become invalid if e.g. the video sink changes, we switch tracks, etc.?

If it can become suddenly invalid, please make this (transfer full) by copying the list and taking another reference of each channel.

lib/gst/player/gstplayer.c
```diff
@@ -3179,6 +3180,84 @@ gst_player_set_visualization_enabled (GstPlayer * self, gboolean enabled)
...
+ */
+const GList *
+gst_player_get_color_balance_list_channels (GstPlayer * self)
+{
+  g_return_val_if_fail (GST_IS_PLAYER (self), NULL);
+
+  if (!GST_IS_COLOR_BALANCE (self->playbin))
```
sdroege: That will always be true :)

lib/gst/player/gstplayer.c
```diff
@@ -3179,6 +3180,84 @@ gst_player_set_visualization_enabled (GstPlayer * self, gboolean enabled)
...
+ * Retrieve the current value of the indicated channel, between min_value
+ * and max_value.
+ *
+ * See Also: The #GstColorBalanceChannel.min_value and
+ *         #GstColorBalanceChannel.max_value members of the
+ *         #GstColorBalanceChannel object.
+ *
+ * Returns: The current value of the channel.
+ */
+gint
```
sdroege: Not sure if these plain gint values are useful here. We should at least document their range, and make sure it's always the same range per channel.


I think in general I would prefer if we don't expose the GstColorBalance API in GstPlayer, but instead have a simple API around an enum { BRIGHTNESS, CONTRAST, SATURATION, HUE } for setting things directly

sdroege:
Thanks for the patch :)

lib/gst/player/gstplayer.c
ceyusa: My doubt was what if a future descendant of GstPlayer uses a not playbin pipeline.

sdroege: Good point. Yet another reason to hide all this GstColorBalance stuff more from the application and do all the difficult things internally :)

ceyusa:
The new API is in place for your review, @sdroege

ceyusa:
@sdroege the new API v3

I added gst_player_has_color_balance()

sdroege:
Thanks, merged with some minor changes :)

PR #58, opened 2015-06-10
misc fixes

ceyusa:
Several fixes minor fixes and improvements.

I still have "build: remove init function in symbol scanners" because the initialization of gstreamer is not required for the scanners, since all the gstreamer api is hidden to the api user.


sdroege:
Merged all but the init function commit :)

ceyusa:
@sdroege for now I give up in removing the gst_init() ;)

But let me insist, at least on the g_type_init().

Also, I found a problem for g-ir-scanner in the annotations.

sdroege:
Thanks :)

PR #60, opened 2015-06-11
player: channel's label is the haystack

ceyusa:
When searching for a specific color balance channel within the ones provided
by the element, the string haystack is the label for the element's channel,
and the string needle is what we are looking.

Thus, we can assign custom color balance elements rather than only playbin.

### DIFF
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -3425,7 +3425,7 @@ gst_player_color_balance_find_channel (GstPlayer * self,
       gst_color_balance_list_channels (GST_COLOR_BALANCE (self->playbin));
   for (l = channels; l; l = l->next) {
     channel = l->data;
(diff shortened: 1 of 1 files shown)


sdroege:
Thanks, merged :) Where are you using GstPlayer, and what kind of non-playbin pipeline do you use there?

ceyusa:
> Thanks, merged :) Where are you using GstPlayer, and what kind of non-playbin pipeline do you use there?

I'm testing new features in vaapipostproc, which I'm working on color balance and skin tone enhancement.

You can see the branch https://github.com/ceyusa/gst-player/commits/vaapipostproc where I search for the vaapipostproc in the pipeline and expose those features in the UI.

sdroege:
Ah, it would be good to integrate that with playbin somehow so it uses vaapipostproc if vaapisink is used. See also https://bugzilla.gnome.org/show_bug.cgi?id=687182 , which is related

PR #62, opened 2015-06-11
gtk-play: remove unused variables

ceyusa:
hihihi :)

### DIFF
--- a/gtk/gtk-play.c
+++ b/gtk/gtk-play.c
@@ -1653,7 +1653,6 @@ gtk_play_constructor (GType type, guint n_construct_params,
     GObjectConstructParam * construct_params)
 {
   GtkPlay *self;
-  gchar **p;
 
   self =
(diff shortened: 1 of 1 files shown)


sdroege:
Thanks :) Why doesn't clang complain about those?!

ceyusa:
> Thanks :) Why doesn't clang complain about those?!

I use clang, and it complains a lot  :)

sdroege:
Why does *my* clang not complain about it :) 3.6.1 btw, what are you using?

ceyusa:
> Why does my clang not complain about it :) 3.6.1 btw, what are you using?

Debian clang version 3.5.2-1 (tags/RELEASE_352/final) (based on LLVM 3.5.2)

The one that debian-testing distributes.

sdroege:
Thanks, I'll try downgrading next time :)

PR #63, opened 2015-06-12
Add playback rate support

codomania:
Initial implementation of playback rate control support. I have modified gtk-play to add new menu option (Playback Speed) to select the playback rates (0.5, 1.0, 1.5, 2.0, 2.5 ... 4.0). I guess in future  we can add a playback rate slider to control the rate instead of menu option..

### DIFF
--- a/gtk/gtk-play.c
+++ b/gtk/gtk-play.c
@@ -1052,6 +1052,46 @@ create_tracks_menu (GtkPlay * play, GstPlayerMediaInfo * media_info, GType type)
 }
 
 static void
+player_set_rate_cb (GtkSpinButton * button, GtkPlay * play)
+{
+  gst_player_set_rate (play->player, gtk_spin_button_get_value (button));
+}
+
+static void
+player_speed_clicked_cb (GtkButton * button, GtkPlay * play)
+{
+  GtkWidget *box;
+  GtkWidget *label;
+  GtkWidget *dialog;
+  GtkWidget *content;
+  GtkWidget *rate_spinbutton;
+
+  dialog = gtk_dialog_new ();
+  gtk_window_set_transient_for (GTK_WINDOW (dialog), GTK_WINDOW (play));
+  gtk_window_set_title (GTK_WINDOW (dialog), "Playback speed control");
+
+  content = gtk_dialog_get_content_area (GTK_DIALOG (dialog));
+  box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 1);
+  gtk_box_pack_start (GTK_BOX (content), box, TRUE, TRUE, 0);
+
+  label = gtk_label_new ("Playback rate");
+  rate_spinbutton = gtk_spin_button_new_with_range (-64, 64, 0.1);
+  gtk_spin_button_set_digits (GTK_SPIN_BUTTON (rate_spinbutton), 2);
+  gtk_spin_button_set_value (GTK_SPIN_BUTTON (rate_spinbutton),
+      gst_player_get_rate (play->player));
+
+  gtk_box_pack_start (GTK_BOX (box), label, TRUE, TRUE, 2);
+  gtk_box_pack_start (GTK_BOX (box), rate_spinbutton, TRUE, TRUE, 2);
+  g_signal_connect (rate_spinbutton, "value-changed",
+      G_CALLBACK (player_set_rate_cb), play);
+
+  gtk_widget_show_all (dialog);
+  gtk_dialog_run (GTK_DIALOG (dialog));
+
+  gtk_widget_destroy (dialog);
+}
+
+static void
 player_quit_clicked_cb (GtkButton * button, GtkPlay * play)
 {
   gtk_widget_destroy (GTK_WIDGET (play));
@@ -1072,6 +1112,7 @@ gtk_player_popup_menu_create (GtkPlay * play, GdkEventButton * event)
   GtkWidget *submenu;
   GtkWidget *vis;
   GtkWidget *cb;
+  GtkWidget *speed;
   GstPlayerMediaInfo *media_info;
 
   menu = gtk_menu_new ();
@@ -1085,6 +1126,7 @@ gtk_player_popup_menu_create (GtkPlay * play, GdkEventButton * event)
   quit = gtk_menu_item_new_with_label ("Quit");
   vis = gtk_menu_item_new_with_label ("Visualization");
   cb = gtk_menu_item_new_with_label ("Color Balance");
+  speed = gtk_menu_item_new_with_label ("Playback Speed");
 
   media_info = gst_player_get_media_info (play->player);
 
@@ -1144,6 +1186,8 @@ gtk_player_popup_menu_create (GtkPlay * play, GdkEventButton * event)
       G_CALLBACK (skip_prev_clicked_cb), play);
   g_signal_connect (G_OBJECT (info), "activate",
       G_CALLBACK (media_info_clicked_cb), play);
+  g_signal_connect (G_OBJECT (speed), "activate",
+      G_CALLBACK (player_speed_clicked_cb), play);
   g_signal_connect (G_OBJECT (quit), "activate",
       G_CALLBACK (player_quit_clicked_cb), play);
 
@@ -1156,6 +1200,7 @@ gtk_player_popup_menu_create (GtkPlay * play, GdkEventButton * event)
   gtk_menu_shell_append (GTK_MENU_SHELL (menu), sub);
   gtk_menu_shell_append (GTK_MENU_SHELL (menu), info);
   gtk_menu_shell_append (GTK_MENU_SHELL (menu), cb);
+  gtk_menu_shell_append (GTK_MENU_SHELL (menu), speed);
   gtk_menu_shell_append (GTK_MENU_SHELL (menu), quit);
 
   gtk_widget_show_all (menu);
--- a/docs/lib/gst-player-sections.txt
+++ b/docs/lib/gst-player-sections.txt
@@ -73,6 +73,9 @@ gst_player_has_color_balance
 gst_player_set_color_balance
 gst_player_get_color_balance
 
+gst_player_get_rate
+gst_player_set_rate
+
 <SUBSECTION Standard>
 GST_IS_PLAYER
 GST_IS_PLAYER_CLASS


lib/gst/player/gstplayer.c
```diff
@@ -2554,6 +2557,7 @@ gst_player_stop_internal (gpointer user_data)
   }
   self->seek_position = GST_CLOCK_TIME_NONE;
   self->last_seek_time = GST_CLOCK_TIME_NONE;
+  self->rate = 1.0;
```
sdroege: Reset trick_play to FALSE here :)

lib/gst/player/gstplayer.c
```diff
@@ -2617,13 +2623,24 @@ gst_player_seek_internal_locked (GstPlayer * self)
   remove_tick_source (self);
   self->is_eos = FALSE;
 
-  ret =
-      gst_element_seek_simple (self->playbin, GST_FORMAT_TIME,
-      GST_SEEK_FLAG_FLUSH, position);
+  flags |= GST_SEEK_FLAG_FLUSH;
+  flags |= GST_SEEK_FLAG_ACCURATE;
```
sdroege: ACCURATE is unnecessarily expensive usually. Did you check if it is really required here?

lib/gst/player/gstplayer.c
```diff
@@ -2617,13 +2623,24 @@ gst_player_seek_internal_locked (GstPlayer * self)
...
 
-  if (!ret)
-    emit_error (self, g_error_new (GST_PLAYER_ERROR, GST_PLAYER_ERROR_FAILED,
+  if (self->trick_play)
+    flags |= GST_SEEK_FLAG_TRICKMODE;
+
+  s_event = gst_event_new_seek (self->rate, GST_FORMAT_TIME, flags,
+      GST_SEEK_TYPE_SET, position, GST_SEEK_TYPE_SET, GST_CLOCK_TIME_NONE);
```
sdroege: If only the rate is changed, we could also consider sending a seek with start_type=GST_SEEK_TYPE_NONE. Elements are supposed to continue from the current position with the new rate then.

Also use GST_SEEK_TYPE_NONE for stop_type here in any case

lib/gst/player/gstplayer.c
```diff
@@ -2641,6 +2658,46 @@ gst_player_seek_internal (gpointer user_data)
...
+
+  self->rate = rate;
+  if (rate != 1.0)
+    self->trick_play = TRUE;
+  else
+    self->trick_play = FALSE;
+
+  self->seek_position = gst_player_get_position (self);
+
+  self->seek_source = g_idle_source_new ();
```
sdroege: Maybe check if a seek_source is already pending here. If so, the next call to gst_player_seek_internal() would already take the latest rate we set and we don't have to send yet another seek
I think we also need to use the mutex here to protect seek_source and seek_position.

lib/gst/player/gstplayer.c
```diff
@@ -2641,6 +2658,46 @@ gst_player_seek_internal (gpointer user_data)
...
+void
+gst_player_set_rate (GstPlayer *self, gdouble rate)
+{
+  g_return_if_fail (GST_IS_PLAYER (self));
+
+  self->rate = rate;
+  if (rate != 1.0)
+    self->trick_play = TRUE;
+  else
+    self->trick_play = FALSE;
```
sdroege: Why not just check for self->rate != 1.0 everywhere instead of having this boolean?

lib/gst/player/gstplayer.c
sdroege: And rate should also be protected by the mutex I guess

sdroege: And maybe also need to do some seek throttling like we do in gst_player_seek() here?

lib/gst/player/gstplayer.c
sdroege: In theory we could also use the step events if the playback direction did not change, they are faster and cheaper. However if the rate changed too much, a seek event would still be useful (as elements can decide to only show keyframes then).

lib/gst/player/gstplayer.c
```diff
@@ -2617,13 +2623,24 @@ gst_player_seek_internal_locked (GstPlayer * self)
...
-  ret =
-      gst_element_seek_simple (self->playbin, GST_FORMAT_TIME,
-      GST_SEEK_FLAG_FLUSH, position);
+  flags |= GST_SEEK_FLAG_FLUSH;
+  flags |= GST_SEEK_FLAG_ACCURATE;
 
-  if (!ret)
-    emit_error (self, g_error_new (GST_PLAYER_ERROR, GST_PLAYER_ERROR_FAILED,
+  if (self->trick_play)
+    flags |= GST_SEEK_FLAG_TRICKMODE;
```
sdroege: Maybe we want to expose this as a property? Selecting what kind of trick mode we want: a) full and accurate, b) allow dropping of non-keyframes, c) drop keyframes, d) disable audio (which actually makes sense together with each of a) b) and c))

lib/gst/player/gstplayer.h
```diff
@@ -83,6 +83,9 @@ void         gst_player_stop                          (GstPlayer    * player);
 
 void         gst_player_seek                          (GstPlayer    * player,
                                                        GstClockTime   position);
+void         gst_player_set_rate                      (GstPlayer    * player,
+                                                       gdouble        rate);
+gdouble      gst_player_get_rate                      (GstPlayer    * player);
```
sdroege: Should probably also have a GObject property for this

lib/gst/player/gstplayer.c
```diff
@@ -2641,6 +2658,46 @@ gst_player_seek_internal (gpointer user_data)
...
+ * gst_player_set_rate:
+ * @player: #GstPlayer instance
+ * @rate: playback rate
+ *
+ * Playback at specified rate
+ */
+void
+gst_player_set_rate (GstPlayer *self, gdouble rate)
+{
+  g_return_if_fail (GST_IS_PLAYER (self));
```
sdroege: rate==0.0 is invalid and should be caught here, or we should convert rate==0.0 to pausing... but that then requires keeping track of much more state everywhere.

sdroege:
Thanks for the patch :) Not all of my comments need to be addressed in a first version btw, some are just bonus and optimizations that can be done later.

codomania:
updated the patch with your review comment. I will add GstSeekFlag property after these patches gets merged.

codomania:
now range is set from -10 to 10. Also updated gtk-play to use slider to set the range.  Did some basic testing with rate control (FF, RR) and seek and it seem to be working okay.

sdroege:
Make it -64 to 64 at least :) People are doing such things ;) Thanks for the update, I'll look at it tomorrow

gtk/gtk-play.c
```diff
@@ -1051,6 +1051,47 @@ create_tracks_menu (GtkPlay * play, GstPlayerMediaInfo * media_info, GType type)
   return menu;
 }
 
+static gboolean
+player_set_rate_cb (GtkRange * range, GtkScrollType scroll,
+    gdouble value, GtkPlay * play)
+{
+  gst_player_set_rate (play->player, value);
```
sdroege: This has to return some boolean

gtk/gtk-play.c
```diff
@@ -1051,6 +1051,47 @@ create_tracks_menu (GtkPlay * play, GstPlayerMediaInfo * media_info, GType type)
...
+static void
+player_speed_clicked_cb (GtkButton * button, GtkPlay * play)
+{
+  GtkWidget *scale;
+  GtkWidget *dialog;
+  GtkWidget *content;
+  GtkWidget *box;
+  gdouble value;
+
+  dialog = gtk_dialog_new ();
```
sdroege: Add some "Close" button to the dialog

gtk/gtk-play.c
```diff
@@ -1051,6 +1051,47 @@ create_tracks_menu (GtkPlay * play, GstPlayerMediaInfo * media_info, GType type)
...
+  gtk_window_set_transient_for (GTK_WINDOW (dialog), GTK_WINDOW (play));
+  gtk_window_set_title (GTK_WINDOW (dialog), "Playback speed control");
+
+  content = gtk_dialog_get_content_area (GTK_DIALOG (dialog));
+  box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 1);
+  gtk_box_pack_start (GTK_BOX (content), box, TRUE, TRUE, 0);
+
+  scale = gtk_scale_new_with_range (GTK_ORIENTATION_HORIZONTAL,
```
sdroege: Maybe we can just put a GtkScaleButton directly into the menu instead of creating a dialog? Or even better a GtkSpinButton directly into the menu.

I think a spin button is also better UI for this in general.


While at it, I think we should also block the seekbar position-update code while the seekbar is dragged. Otherwise if you slowly drag it around, it will jump back and forth all the time between the position it just seeked to and the position the mouse drags it to

sdroege:
I merged the non-GTK change already and cleaned it up a little

gtk/gtk-play.c
codomania: I did thought about it but then inserting slider or spinbutton inside the menu itself will create usability and accessibility problems.  e.g it would interface with expected  keyboard navigation behavior when add text into text spinbox.  Those are some reason behind using dialog box. In general i think our gui controls placement needs a bit rework. Since now our feature list are growing hence it maybe good idea to just revisit on what goes where.. so far we just took a easy path on putting all the stuff inside either popup menu item or created button in control panel ;)

codomania: i saw @amorenoc nice look GUI, maybe once that is completed then we can borrow the idea from it and then make similar looks and feel modification in gtk app...

sdroege: Yes, that's my "vision" for the GTK UI too. Similar to how snappy ( https://wiki.gnome.org/Apps/Snappy/ ) looks, but without depending on clutter.

sdroege: The slider is also not very usable IMHO, maybe a dialog with a spinbutton then? Just sliding the playback speed is a bit weird :)

codomania: sure will go with spin button :)  actually slider was inspired from vlc playrate control.

codomania: just downloaded snappy and main UI looks pretty nice.. looks like if we like to do similar in gtk then it may require a good amount of cairo draw code to get similar looks and feel. It seems most of those advance controls  (i.e stream selection, media info, rate control, loop etc) are done via keyboard event.

sdroege:
Thanks, merged :)

PR #64, opened 2015-06-14
Add more unit test

codomania:
I have transcoded sintel clip and add some unit test to test the following
1) media info (video, audio and subtitle)
2) stream switching 
3) stream enable/disable

Stream enable/disable is not working well. I will submit this as bug in gstreamer database.


tests/player.c
```diff
@@ -32,6 +32,7 @@
 #endif
 
 #include <check.h>
+#include <unistd.h>
```
sdroege: Why?

tests/player.c
```diff
@@ -536,6 +439,335 @@ START_TEST (test_play_audio_eos)
...
+  test_video_info (media_info);
+}
+
+static void
+test_play_media_info_cb (GstPlayer * player, TestPlayerStateChange change,
+    TestPlayerState * old_state, TestPlayerState * new_state)
+{
+  if (change == STATE_CHANGE_MEDIA_INFO_UPDATED)
+    test_media_info_object (player, new_state->media_info);
+  else if (change == STATE_CHANGE_END_OF_STREAM)
```
sdroege: Let's not wait until EOS but after it had media info, or on ERROR/EOS or playing a while without any mediainfo

tests/player.c
```diff
@@ -536,6 +439,335 @@ START_TEST (test_play_audio_eos)
...
+
+  fail_unless (player != NULL);
+
+  uri = gst_filename_to_uri (TEST_PATH "/sintel.mkv", NULL);
+  fail_unless (uri != NULL);
+  gst_player_set_uri (player, uri);
+  g_free (uri);
+
+  gst_player_play (player);
+  g_main_loop_run (state.loop);
```
sdroege: You should check after g_main_loop_run() that you actually got media info callback

tests/player.c
```diff
@@ -536,6 +439,335 @@ START_TEST (test_play_audio_eos)
...
+    gint fps_d, fps_n;
+    guint par_d, par_n;
+    GstPlayerStreamInfo *stream = (GstPlayerStreamInfo *) list->data;
+    GstPlayerVideoInfo *video_info = (GstPlayerVideoInfo *) stream;
+
+    fail_unless (gst_player_stream_info_get_tags (stream) != NULL);
+    fail_unless (gst_player_stream_info_get_caps (stream) != NULL);
+    fail_unless_equals_int (gst_player_stream_info_get_index (stream), 0);
+    fail_unless_equals_string (gst_player_stream_info_get_codec (stream),
+        "H264");
```
sdroege: Fails with latest GStreamer GIT:
player.c:498:F:general:test_play_media_info:0: 'gst_player_stream_info_get_codec (stream)' (H.264 (High Profile)) is not equal to '"H264"' (H264)

tests/player.c
```diff
@@ -536,6 +439,335 @@ START_TEST (test_play_audio_eos)
...
+  gint completed = GPOINTER_TO_INT (new_state->test_data);
+
+  /* If pipeline is playing and stream switching is not tested
+   * then switch to next available streams */
+  if (new_state->state == GST_PLAYER_STATE_PLAYING && !completed) {
+    switch_streams (player, GST_TYPE_PLAYER_SUBTITLE_INFO);
+    switch_streams (player, GST_TYPE_PLAYER_AUDIO_INFO);
+    new_state->test_data = GINT_TO_POINTER (completed + 1);
+  }
+  else if (change == STATE_CHANGE_END_OF_STREAM)
```
sdroege: EOS here is also one minute away, that's a bit long if not really needed. Stop after all stream switches were successful

tests/player.c
```diff
@@ -536,6 +439,335 @@ START_TEST (test_play_audio_eos)
...
+      } else if (type == GST_TYPE_PLAYER_AUDIO_INFO) {
+        fail_unless (gst_player_set_audio_track (player, index) == TRUE);
+        new_stream = (GstPlayerStreamInfo *)
+                          gst_player_get_current_audio_track (player);
+      }
+
+      fail_unless_equals_int (gst_player_stream_info_get_index (new_stream),
+                              index);
+      /* wait for 2 seconds before switching to next stream */
+      sleep (2);
```
sdroege: Don't sleep, use some of the GMainContext API to just call your callback again at a later time. g_timeout_add() for example.

tests/player.c
```diff
@@ -536,6 +439,335 @@ START_TEST (test_play_audio_eos)
...
+    TestPlayerState * new_state)
+{
+  gint completed = GPOINTER_TO_INT (new_state->test_data);
+
+  /* If pipeline is playing then disable all the streams */
+  if (new_state->state == GST_PLAYER_STATE_PLAYING && !completed) {
+    disable_all_streams (player);
+    new_state->test_data = GINT_TO_POINTER (completed + 1);
+  }
+  else if (change == STATE_CHANGE_END_OF_STREAM)
```
sdroege: And here the same about EOS

sdroege:
Thanks for the patches :) There are some known problems with stream switches, which one did you observe here? I didn't have the patience to let it run until EOS ;)

tests/player.c
codomania: one question, the reason behind sleep was that I was planning to do something like this in test:

- save the current player position (cur_position)
- switch to new stream 
- sleep for 3 seconds 
- query the current position again 
- compare the new current position with previous position and if the position is not changed then 
pipeline is stalled somewhere.

Is it okay to call sleep() in this case ?  I have ran into instances where the stream is switched but then it get stalled during playback. If we simply query the current stream index then it will give us wrong data. Of course,  if you wait for 2 min then tcase will timeout and that will trigger failure ;) 

Let me know your thought

sdroege: Asynchronous waiting via the mainloop is better in any case, even if it's quite inconvenient in C unfortunately :(

sdroege: Problem is that with sleep() you're going to block the mainloop for the whole time

codomania:
Thanks for review, patch set are updated with your review comments.

tests/player.c
sdroege: On errors and without getting any media-info, you would still succeed this test here

tests/player.c
```diff
@@ -536,6 +439,451 @@ START_TEST (test_play_audio_eos)
...
+  fail_unless (gst_player_set_subtitle_track
+                (args->player, args->index) == TRUE);
+  new_stream = (GstPlayerStreamInfo *)
+                gst_player_get_current_subtitle_track (args->player);
+  fail_unless_equals_int (
+                gst_player_stream_info_get_index (new_stream),
+                args->index);
+
+  args->new_state->test_data = GINT_TO_POINTER (step + 1);
+  return FALSE;
```
sdroege: G_SOURCE_REMOVE is more meaningful than FALSE

tests/player.c
```diff
@@ -536,6 +439,451 @@ START_TEST (test_play_audio_eos)
...
+                (args->player, args->index) == TRUE);
+  new_stream = (GstPlayerStreamInfo *)
+                gst_player_get_current_audio_track (args->player);
+
+  fail_unless_equals_int (
+                gst_player_stream_info_get_index (new_stream),
+                args->index);
+
+  args->new_state->test_data = GINT_TO_POINTER (step + 1);
+  return FALSE;
```
sdroege: G_SOURCE_REMOVE is more meaningful than FALSE

tests/player.c
```diff
@@ -536,6 +439,451 @@ START_TEST (test_play_audio_eos)
...
+}
+
+static gboolean
+switch_streams_completed_loop (gpointer user_data)
+{
+  TestPlayerState *state = (TestPlayerState *) user_data;
+  gint step = GPOINTER_TO_INT (state->test_data);
+
+  if (step != 10)
+    return TRUE;
```
sdroege: G_SOURCE_CONTINUE

tests/player.c
```diff
@@ -536,6 +439,451 @@ START_TEST (test_play_audio_eos)
...
+
+  /* if all the stream switching is completed then quit the loop */
+  g_main_loop_quit (state->loop);
+  return FALSE;
```
sdroege: G_SOURCE_REMOVE

tests/player.c
```diff
@@ -536,6 +439,451 @@ START_TEST (test_play_audio_eos)
...
+
+  /* If pipeline is playing and stream switching thread is not started
+   * then switch to next available streams */
+  if (new_state->state == GST_PLAYER_STATE_PLAYING && !running) {
+    GstPlayerMediaInfo *media_info;
+
+    media_info = gst_player_get_media_info (player);
+    switch_streams (player, gst_player_get_subtitle_streams (media_info), new_state);
+    switch_streams (player, gst_player_get_audio_streams (media_info), new_state);
+    g_idle_add (switch_streams_completed_loop, new_state);
```
sdroege: You have to copy the state, it will continue to be updated

tests/player.c
```diff
@@ -536,6 +439,451 @@ START_TEST (test_play_audio_eos)
...
+    gint fps_d, fps_n;
+    guint par_d, par_n;
+    GstPlayerStreamInfo *stream = (GstPlayerStreamInfo *) list->data;
+    GstPlayerVideoInfo *video_info = (GstPlayerVideoInfo *) stream;
+
+    fail_unless (gst_player_stream_info_get_tags (stream) != NULL);
+    fail_unless (gst_player_stream_info_get_caps (stream) != NULL);
+    fail_unless_equals_int (gst_player_stream_info_get_index (stream), 0);
+    fail_unless (strstr (gst_player_stream_info_get_codec (stream),
+        "H264") != NULL);
```
sdroege: It's "H.264 (High Profile)" here, note the dot

tests/player.c
codomania: Actually in callback, I am using loop and test_data variables and based on the condition will either increment the step or quit the loop.  I believe these variables are not updated in player cb. But if recommendation is to create a copy then sure I can do that. let me know your thought. thanks

sdroege: The new_state can be modified between g_idle_add() and when your callback is called. So copy :)

codomania:
Thanks for review, the patch set is updated.

sdroege:
Thanks, merged :)

Issue #65, opened 2015-06-17
external subtitle unit test


codomania:
FYI, i was working on developing external subtitle unit test and found some minor issue with handling invalid suburis. I have traced down the issue in gst-plugins-base https://bugzilla.gnome.org/show_bug.cgi?id=751118. Once this gets fixed in upstream then we may also need to make some smaller changes in our libs to reflect it. e.g currently we are returning self->suburi when user queries the current suburi but this need to be changed so that we query the suburi from playbin's "current-suburi" property to get the correct state.

sdroege:
Thanks, I'll take a look at that bug in the next days. Weird that nobody else noticed that yet :)

sdroege:
This should be ok now

PR #66, opened 2015-06-17
Add external subtitle uri and visualization unit test

codomania:
currently invalid suburi unit test is failing, see #65 for details.


lib/gst/player/gstplayer.c
```diff
@@ -515,12 +511,16 @@ gst_player_get_property (GObject * object, guint prop_id,
...
+
       g_mutex_lock (&self->lock);
-      g_value_set_string (value, self->suburi);
+      g_object_get (G_OBJECT (self->playbin), "current-suburi", &suburi, NULL);
+      g_value_set_string (value, suburi);
       g_mutex_unlock (&self->lock);
-      GST_DEBUG_OBJECT (self, "Returning has-suburi=%d",
-          g_value_get_boolean (value));
+      GST_DEBUG_OBJECT (self, "Returning has-suburi=%d", suburi ? TRUE : FALSE);
+      g_free (suburi);
```
sdroege: I don't like that we use self->uri for the URI property but playbin's current-suburi for the sub URI property. That seems inconsistent and we should make that more unified

tests/player.c
```diff
@@ -883,6 +891,478 @@ START_TEST (test_play_stream_disable_enable)
...
+    new_state->test_data = GINT_TO_POINTER (running + 1);
+
+    suburi = gst_filename_to_uri (TEST_PATH "/foo.srt", NULL);
+    fail_unless (suburi != NULL);
+
+    /* load invalid suburi */
+    fail_unless (gst_player_set_subtitle_uri (player, suburi) != FALSE);
+    g_free (suburi);
+
+    g_timeout_add_full (G_PRIORITY_DEFAULT, 5000, invalid_external_suburi_test,
```
sdroege: What are you waiting for here for 5 seconds?

sdroege: I mean, add a comment :)

sdroege:
Looks generally good, thanks

lib/gst/player/gstplayer.c
codomania: yes agree, we can take care of this in next patch.

tests/player.c
codomania: sure will add comment. the reason behind waiting is that set_subtitle uri will internally destroy the pipeline, create a new pipeline and also set the state etc and this can take sometime hence querying the get_subtitle_uri() just after set_subtitle_uri() will not give us accurate status.

sdroege: Can we know more about this somehow? Via GstPlayer state changes maybe, or some warning message?

codomania: playbin does generate warning when it fails to load the suburi and our warning_cb does get called appropriately. Maybe we should install the warning cb handler in app and take action on that warning.

codomania: i just looked at gstplayer and it seems like we do not emit warning messages to the app. so to use the warning method we first need to add the emit logic.

codomania:
okay added support to emit warning signal and updated test to use that signal to verify invalid suburi. see if it make sense

codomania:
emit warning signal patch is updated with your review feedback.

sdroege:
I've merged the warning signal, some comments for the other things :)

codomania:
I have dropped vis unit test for now as you pointing that we should probably consider adding property in player to get/set/current vis elements, this will simplify our test as well as usage outside the test. 

have updated the rate unit test to remove the g_timeout loop and now checking the position. The forward rate is working just fine but for backward rate is failing for us. Not sure if its test issue or gstplayer issue at this time.

codomania:
now use media info signal instead of 5 second wait time. I will look at implementing similar logic in stream selection unit test to remove 5 second wait logic.

sdroege:
Great, thanks :) I'll look at merging these changes tonight

sdroege:
For the vis test, you could probably also check each time the position-updated signal is emitted if we now switched to the new vis element. That's not perfect but you would still only wait as long as really needed (and at most 100ms longer).

sdroege:
Thanks, merged all that :) Now we should fix the backwards test at some point

PR #68, opened 2015-06-27
gtk-play: build toolbar from xml and css

codomania:
Added support to build the UI from glade generated xml files and also added support to apply css. This should help us to customize the looks and feel without making too much changes in C code. I am not UI designer by any stretch but here is what I was quickly able to customize. The toolbar now looks like this:

screenshot from 2015-06-26 19 29 11

The code is fully function, also did some cleanup around the image area widget. Now toolbar is overlay on top of video which also let us get those cool transparent widget. If this looks good then next i was going to look into adding app menu (again possibly generate GMenu through glade) and move the functionality of popup into app menu. Let me know what you think  ?

Cleanup includes:
* remove the image_area widget - the main purpose of this was to test media info provided cover image logic and we know it works well. Now instead of display image in window, i simply set the pixbuf in title. So if window manager respects the title image then you should see the image in title.
* remove visual args. Now visualization is enabled by default and if you play audio then visualization will automatically enabled.
* For now remove the loop button, will find a new place of it once we add app menu support. But one can still enable the loop through command line args (e.g --loop flags)
* media info dialog is also generated from the xml. I was not able to get the treemodel working through glade hence part of media info is still done through C code. I will later look into it or if someone knows the glade ticks to make the treeview work then it will be really cool as it can help remove some more code.
* Autohide toolbar no longer need to be done when we are in fullscreen. The toolbar is auto hidden even in  normal mode (inspired from snappy)
* toolbar also prints label when playing > or < 1.0 rate.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -60,8 +60,11 @@ AC_SUBST(GLIB_PREFIX)
 GST_PREFIX="`$PKG_CONFIG --variable=prefix gstreamer-1.0`"
 AC_SUBST(GST_PREFIX)
 
-PKG_CHECK_MODULES(GTK, [gtk+-3.0], [have_gtk="yes"], [have_gtk="no"])
+PKG_CHECK_MODULES(GTK, [gtk+-3.0 >= 3.14], [have_gtk="yes"], [have_gtk="no"])
 AM_CONDITIONAL(HAVE_GTK, test "x$have_gtk" != "xno")
+
+PKG_CHECK_MODULES(GMODULE, [gmodule-2.0], [have_gmodules="yes"], [have_modules="no"])
+AM_CONDITIONAL(HAVE_GMODULE, test "x$have_gmodules" != "xno")
 
 PKG_CHECK_MODULES(GTK_X11, [gtk+-x11-3.0 x11], [have_gtk_x11="yes"], [have_gtk_x11="no"])
 AM_CONDITIONAL(HAVE_GTK_X11, test "x$have_gtk_x11" != "xno")
--- a/gtk/Makefile.am
+++ b/gtk/Makefile.am
@@ -1,10 +1,36 @@
 bin_PROGRAMS = gtk-play
 
-gtk_play_SOURCES = gtk-play.c
+gtk-play-resources.c: resources/gresources.xml \
+	resources/media_info_dialog.ui	\
+	resources/toolbar.css						\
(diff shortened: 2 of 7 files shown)


sdroege:
Oh that already looks like a great improvement :)

gtk/Makefile.am
```diff
@@ -1,10 +1,30 @@
 bin_PROGRAMS = gtk-play
 
-gtk_play_SOURCES = gtk-play.c
+gtk-play-resources.c:	$(shell find resources )
```
sdroege: This should probably explicitly list the resources, also below

gtk/resources/toolbar.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+GtkButton:hover {
+  border-radius: 45px;
+  background: rgba(43,56,54,1.0);
+  border: 1px solid #212B2A;
+}
+
+#title_label {
+  font: Comic San 10;
```
sdroege: Comic Sans? :)

sdroege:
Two things I noticed:
- when being over the toolbar and stopping to move the mouse, it also disappears. It probably should only disappear if the mouse it not moved outside the toolbar
- the toolbar sometimes also disappears when the mouse is continued to be moved over the toolbar
- seeking only happens when releasing the mouse button now. It would be nicer if it would seek continuously while moving the slider (and stop updating the slider from the position signal while the mouse button is down)

Otherwise really great :)

configure.ac
```diff
@@ -63,6 +63,9 @@ AC_SUBST(GST_PREFIX)
 PKG_CHECK_MODULES(GTK, [gtk+-3.0], [have_gtk="yes"], [have_gtk="no"])
```
sdroege: This probably needs a newer GTK, do you know which version at least?

codomania:
updated commit with review comment. regarding the gtk version i have "3.14.13-0ubuntu1", I believe we need a min of 3.10+. But again i am not sure when the css finally got added in gtk. 
let me know if you want me to restrict to version 3.14 ?

sdroege:
3.14 sounds ok, if someone runs something smaller than that then they can check if it builds and works and update configure.ac ;)

codomania:
updated configure.ac with gtk+-3.0 >= 3.14 check.

sdroege:
Thanks, merged :)

PR #73, opened 2015-07-10
Emit seek-done signal when the seek is finished.

aobzhirov:
We added "seek-done" event which we need for Chromium GStreamer backend.

### DIFF
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -96,6 +96,7 @@ enum
   SIGNAL_MEDIA_INFO_UPDATED,
   SIGNAL_VOLUME_CHANGED,
   SIGNAL_MUTE_CHANGED,
+  SIGNAL_SEEK_DONE,
   SIGNAL_LAST
 };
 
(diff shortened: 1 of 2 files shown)


lib/gst/player/gstplayer.c
```diff
@@ -1370,6 +1376,8 @@ state_changed_cb (GstBus * bus, GstMessage * msg, gpointer user_data)
           gst_player_seek_internal_locked (self);
         } else {
           GST_DEBUG_OBJECT (self, "Seek finished");
+          g_signal_emit (self, signals[SIGNAL_SEEK_DONE], 0,
+              gst_player_get_position (self));
```
sdroege: See how the other signals are emitted, you need to dispatch it to the application's main context if that was requested by the application.

sdroege:
I think this makes sense, just needs a minor fix.


I guess at some point I should find the time to make the signal dispatching a bit nicer. Plan is to let GstPlayer just call a vfunc for emitting signals (which by default just calls g_signal_emit()) and then additionally have subclasses that override the vfunc to dispatch to a GLib main context, or Qt main loop, or .... At the same time we could also add platform-specific subclasses with appropriate video sinks (e.g. a GtkPlayer thing that always uses gtk*sink, the same thing for Qt and Cocoa/OSX).

lib/gst/player/gstplayer.c
aobzhirov: Thanks, will make a change.

sdroege: Great, let me know if you have further questions :)

aobzhirov:
Updated.

sdroege:
Thanks, merged

PR #75, opened 2015-07-17
handle key strokes for controlling media

ceyusa:


gtk/gtk-play.c
```diff
@@ -207,6 +207,58 @@ gtk_play_set_rate (GtkPlay * play, gdouble step)
...
+    case GDK_KEY_Escape:
+      gtk_toggle_button_set_active (GTK_TOGGLE_BUTTON (play->fullscreen_button),
+          FALSE);
+    case GDK_KEY_Right:
+      if (g_list_next (play->current_uri))
+        gtk_button_clicked (GTK_BUTTON (play->next_button));
+      break;
+    case GDK_KEY_Left:
+      if (g_list_previous (play->current_uri))
+        gtk_button_clicked (GTK_BUTTON (play->prev_button));
```
sdroege: Maybe it would be better to use the same keybindings as other media players (totem and snappy and mplayer for example) use, e.g. cursor keys for seeking.

sdroege:
Thanks, that's something I always wanted to add but got distracted so far :)

sdroege:
Are you planning to update the commit to use the same keybindings as other software?

ceyusa:
> Are you planning to update the commit to use the same keybindings as other software?

I'm thinking of mplayer keybindings. But right now I can't work on it. Perhaps this weekend. If you want to tackle it  before, go ahead :)

ceyusa:
Done! What do you think, sdroege?

sdroege:
Looks good except for the rate related bug :)

sdroege:
Merged

PR #76, opened 2015-07-20
misc fixes

ceyusa:
duration and position value fix
subtitle uri fix

### DIFF
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -3371,7 +3371,6 @@ gboolean
 gst_player_set_subtitle_uri (GstPlayer * self, const gchar * suburi)
 {
   g_return_val_if_fail (GST_IS_PLAYER (self), FALSE);
-  g_return_val_if_fail (suburi != NULL, FALSE);
 
   g_mutex_lock (&self->lock);
   if (self->suburi)
--- a/lib/gst/player/gstplayer.c
+++ b/lib/gst/player/gstplayer.c
@@ -553,7 +553,7 @@ gst_player_get_property (GObject * object, guint prop_id,
           g_value_get_boolean (value));
       break;
     case PROP_POSITION:{
-      gint64 position;
+      gint64 position = 0;
 
       gst_element_query_position (self->playbin, GST_FORMAT_TIME, &position);
       g_value_set_uint64 (value, position);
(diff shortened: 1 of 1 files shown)


sdroege:
Thanks, merged

codomania:
sorry for replying a bit late, Regarding second patch (a0cca75), i thought below line in player take care of setting subtitle to NULL when changing the uri, does it not work ? 

https://github.com/sdroege/gst-player/blob/fd672985acf35c9bd4991f83dacb899053e47097/lib/gst/player/gstplayer.c#L421

ceyusa:
> sorry for replying a bit late, Regarding second patch (a0cca75), i thought below line in player take care of setting subtitle to NULL when changing the uri, does it not work

mmmhh... I didn't pay attention to that part. I didn't debugged it, but for sure it didn't work for me.

I set the external suburi right after the uri, but when I changed the media uri (without changing the state) the new video was rendered with the old subtitles, I had to set them explicitly to NULL.

codomania:
Is this something you can reproduce in gtk-play ?  IIRC, we tested this usecase but its possible that its broken now. which version of gstreamer you are using ?  maybe we should consider adding this usecase in test suite.  I have TODO item to remove suburi and uri local variable from GstPlayer instance. The idea is when application asks we should simply set/get playbin's suburi and uri property.

codomania:
Nope you do not need to pause pipeline to set subtitle. I believe the current gtk-play works as well

1) gtk-play sets the subtitle uri.
2) gst-player internally sets the suburi and seeks the pipeline to its current position
3) If pipeline was in pause state then gtk-play sets the pipeline to pause (gst_player_pause)
4) if pipeline was playing state then set the pipeline to (gst_player_play).

Take a look at gtk-play (i think its play_current_uri function).

ceyusa:
> Nope you do not need to pause pipeline to set subtitle. I believe the current gtk-play works as well
>
> 1) gtk-play sets the subtitle uri.
> 2) gst-player internally sets the suburi and seeks the pipeline to its current position

https://github.com/sdroege/gst-player/blob/master/lib/gst/player/gstplayer.c#L444

I was wrong, not pause but ready: gst-player sets the pipeline to READY 

> 3) If pipeline was in pause state then gtk-play sets the pipeline to pause (gst_player_pause)
> 4) if pipeline was playing state then set the pipeline to (gst_player_play).
> 
> Take a look at gtk-play (i think its play_current_uri function).

The problem (for me) of gtk-play UX, is that the user needs to load the external subtitles when the video is already playing. I want to "auto-select" them:

https://github.com/ceyusa/dplayer/blob/master/dmp-player.c#L518

sdroege:
There's a bug in GStreamer bugzilla with an initial patch by me from many years ago to add external subtitle selection during PLAYING support to playbin. It needed much further work to be useful though. Maybe you want to pick up that one? https://bugzilla.gnome.org/show_bug.cgi?id=589515

PR #80, opened 2015-07-30
gtk-play: fix elapsed and remain time label

codomania:
If elapsed or remain time is greater than 60 mins then add hours in
label text.


codomania:
second commit is just cosmetic fix in toolbar title label. If uri is long then it expands outside the toolbar title label and by using the basename we will restrict it.

sdroege:
Thanks, merged

PR #81, opened 2015-08-15
player: Fix wrong return value warnings

joybro:
error: 'return' with no value, in function returning non-void [-Werror]


sdroege:
Thanks, merged :)

PR #83, opened 2015-08-18
[iOS] Fix Build Issue: Undefined symbols for architecture i386

TrungDoan:
Fix build issue: “Undefined symbols for architecture i386” on iOS by adding additional files in lib/gst/player to
the iOS project.


sdroege:
Thanks, merged :)

PR #100, opened 2015-08-26
gst-play: Add a check condition in print_all_stream_info

JimmyOhn:
I add a check condition for stream_list in print function.
In another print function, it also check whether it's null or not.

### DIFF
--- a/gst-play/gst-play.c
+++ b/gst-play/gst-play.c
@@ -195,9 +195,12 @@ static void
 print_all_stream_info (GstPlayerMediaInfo * media_info)
 {
   guint count = 0;
-  GList *l, *list;
+  GList *list = NULL, *l;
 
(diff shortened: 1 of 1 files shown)


gst-play/gst-play.c
```diff
@@ -195,9 +195,12 @@ static void
 print_all_stream_info (GstPlayerMediaInfo * media_info)
 {
   guint count = 0;
-  GList *l, *list;
+  GList *list = NULL, *l;
```
sdroege: No need to initialize it with NULL here. gst_player_media_info_get_stream_list() will always return something

sdroege:
Looks mostly good, just a code style issue :) Please update

JimmyOhn:
I attached the new commit for your comment.

sdroege:
Thanks, merged

PR #101, opened 2015-09-01
remove unused variable

amorenoc:


sdroege:
Thanks, merged :)

How's your progress with the Qt player? Anything new there?

amorenoc:
I started last weekend to implement the new interfaces, very likely will have something this weekend.

amorenoc:
@sdroege added new interfaces, but having some problems. One the main problems is that I can't find/use the new qml sink, at least not directly from release 1.5.90. Checking the code looks to me that headers aren't installed, and can't use the QuickItem there.

sdroege:
Yes we don't build the Qt sink with the binaries at this point because Qt is huge. You would have to build it yourself from source

PR #102, opened 2015-09-02
gtk-play: Modify switch statement in stream_info_get_string

JimmyOhn:
In stream_info_get_string function, buffer variable is used all switch case repeatedly. Also, return used in switch statement. So, I have modified that cases.

### DIFF
--- a/gst-play/gst-play.c
+++ b/gst-play/gst-play.c
@@ -197,10 +197,6 @@ print_all_stream_info (GstPlayerMediaInfo * media_info)
   guint count = 0;
   GList *list, *l;
 
-  list = gst_player_media_info_get_stream_list (media_info);
-  if (!list)
-    return;
-
   g_print ("URI : %s\n", gst_player_media_info_get_uri (media_info));
   g_print ("Duration: %" GST_TIME_FORMAT "\n",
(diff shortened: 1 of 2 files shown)


gst-play/gst-play.c
```diff
@@ -195,9 +195,12 @@ static void
 print_all_stream_info (GstPlayerMediaInfo * media_info)
 {
   guint count = 0;
-  GList *l, *list;
+  GList *list, *l;
 
   list = gst_player_media_info_get_stream_list (media_info);
+  if (!list)
+    return;
```
sdroege: Even if we have no stream list here, we could at least print the URI, duration and other stream-list-independent things below.

sdroege:
Can you update your pull request?

JimmyOhn:
Sorry for my late reply. I got back from vacation. I'll update ASAP.

sdroege:
Sure, no worries :)

JimmyOhn:
I have modified above code for your comment.:)

sdroege:
Looks good now but can you squash together the commits? One for the list stuff in gst-play, another for the gtk-play parts?

JimmyOhn:
I have modified commits:)

sdroege:
Thanks, merged

PR #105, opened 2015-09-17
gtk-play: change to use valid macro function

zzoon:
non-void function 'gst_player_gtk_video_renderer_get_widget' should return a value


sdroege:
Thanks, merged :)

PR #106, opened 2015-09-18
fix code style

ceyusa:
It seems that, with the long variable names, the gst-indent gets confused and
breaks wrongly some code lines.

This patch aims to a rather sane code style, even if the code lines are longer
than 80 chars.

### DIFF
--- a/gtk/gtk-video-renderer.c
+++ b/gtk/gtk-video-renderer.c
@@ -33,9 +33,7 @@ struct _GstPlayerGtkVideoRendererClass
   GObjectClass parent_class;
 };
 
-static void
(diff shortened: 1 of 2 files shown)


sdroege:
The problem with this is that it will all be reverted again next time someone runs gst-indent

ceyusa:
> The problem with this is that it will all be reverted again next time someone runs gst-indent

What would be the fix? Use another gst-indent script? gst-player-indent ? :)

sdroege:
Modernize gst-indent :)

sdroege:
Let's close this here for now :) Should be tracked here https://bugzilla.gnome.org/enter_bug.cgi?product=GStreamer

PR #110, opened 2015-10-01
qt bindings and qml player

amorenoc:
I believe the current player is in good shape for a merge, have a look when you have time, thanks :)
There are quite a few APIs haven't been added yet, but the basics are there.

screen shot 2015-10-02 at 02 15 00


sdroege:
Looks great! :) Can you squash all the commits in the qt directory together to a single commit (like "Add Qt bindings and player")? If you have any changes elsewhere, those should be separate commits.

Also I think it would be good to have the Qt bindings and the player application separated somehow, different directories or something :)

sdroege:
FWIW, I think for a first version having everything in one directory is ok too! Did you try this only on OSX or also on other platforms?

amorenoc:
Just OSX so far, but I may try it on Linux when I have time. Ok I will squash the commits first. Thanks!

sdroege:
Can you add some license foo at the top of your newly added source file (.h and .cpp is probably enough)? Similar to what the other files already have :)


It also works on Linux btw, just QML seems to have problems in general with hidpi displays.

amorenoc:
Updated PR :+1:

sdroege:
Great, thanks :) I'll try to get through this asap. Why do you ship your own font btw?

amorenoc:
I am using Font Awesome for the icons, we could change that to download it from a CDN, which actually QML FontLoader supports out of the box.

sdroege:
Then you would need internet access to use the application, bundling seems better then :) What is the license of the font?

amorenoc:
SIL Open Font License (OFL), listed here http://www.gnu.org/licenses/license-list.html

sdroege:
Sounds good, can you add a text file next to the font with that information? Where it is from and the license. Maybe just same filename as the font, but .txt in the end?

sdroege:
Thanks, merged!

PR #115, opened 2015-10-22
player: Remove check condition for g_free function

JimmyOhn:
As I know, g_free function is no need to check whether parameter is NULL or not. so I've remove it.


sdroege:
Thanks, merged!

PR #128, opened 2015-10-27
qt: consistent use of pixel sizes. Fixes #124

### LINKED ISSUE #124: qt: Does not work well with HiDPI displays on Linux
@amorenoc see attached screenshot. How does it work on OSX with Retina displays?

It looks like parts of the UI are scaled properly, other parts not.
qt

amorenoc:


sdroege:
While that makes it consistent, I think it is exactly the wrong way around. Now the UI is super small on HiDPI displays :) I think you should use points instead of pixels everywhere, as points are the "virtual" unit of measurement on HiDPI displays.

amorenoc:
I agree, but the correct way is not trivial to implement, although certainly doable :) Will get back to this once we make the UI less pixel-dependent, there are good guidelines for that in:
* http://doc.qt.io/qt-5/scalability.html 
* https://community.kde.org/KDE/High-dpi_issues and
* http://v-play.net/2014/11/supporting-multiple-screen-sizes-and-screen-densities-with-qt-and-v-play/

sdroege:
What's needed for that? Can't you just define the unit of all pixel values there currently, and make them device independent?

sdroege:
I merged this commit now btw, it's better than what we had before in any case :)

PR #134, opened 2015-11-01
qt: seek while dragging seek bar

### LINKED ISSUE #131: qt: Seek bar is updated for position changes while dragging it
@amorenoc When dragging the seek bar (i.e. clicking on it, not releasing the mouse button and moving it around), the seek bar will jump around because of the position updates.

While the seek bar is dragged, it should not be updated from position updates.

amorenoc:
removed time label on top the seek bar, should be placed elsewhere
Fixes #131


sdroege:
Thanks, merged

PR #135, opened 2015-11-01
qt: accept movies as command line parameters

### LINKED ISSUE #122: qt: Does not accept movies as commandline parameters
@amorenoc It should probably be possible to call it with a list of movie files from the commandline.

Also when doing that, it should start playing the first one immediately. And also start playing immediately when a movie is selected from the file chooser (or at least go to paused in both cases so you can see the first frame).

amorenoc:
resolve #122

### DIFF
--- a/qt/main.cpp
+++ b/qt/main.cpp
@@ -20,12 +20,29 @@
 
 #include <QApplication>
 #include <QQmlApplicationEngine>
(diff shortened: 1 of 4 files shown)


sdroege:
Merged but this doesn't completely fix issue #122 , will put details there :)

PR #136, opened 2015-11-01
qt: add new qml item to render media info sample

amorenoc:
when video is not available it will try to display the sample image
returned by media info object.

screen shot 2015-11-02 at 01 53 37


sdroege:
Thanks, merged

PR #147, opened 2016-02-05
android: Move to gradle based build

ford-prefect:


sdroege:
See https://github.com/sdroege/gst-launch-remote/pull/3#issuecomment-184343761

PR #149, opened 2016-03-16
Changing True to glib TRUE to make gtk player project compile

mawasak:
Hi,
To compile gtk-play I have to change True to TRUE.
Regards,
Mariusz


sdroege:
Thanks, merged! Not sure what defined True for me

PR #160, opened 2016-08-08
gtk-play: provide similar behaviour for quit and close

maxinbjohn:
In x86 targets, gtk-play just pause rather than quitting the application
when we click the close button (delete-event). Change the callback function
to get similar behaviour when we click on "Quit" menu option.

Signed-off-by: Maxin B. John <maxin.john@intel.com>


sdroege:
Thanks, merged

PR #161, opened 2016-08-23
android: Update build tools and gradle

ford-prefect:
New build tools needed for Instant Run, and Android Studio wants us to
update that and the gradle version.


sdroege:
Thanks, merged

PR #162, opened 2016-09-02
Universal build + update to Nougat

ford-prefect:
Wait for universal builds to actually become available, of course. :)


sdroege:
Why update to Nougat? :)

ford-prefect:
Mostly because latest and greatest. Don't know if there's a downside to it. Also I could only test against Marshmallow.

